PR #3, opened 2021-01-06
Add event model protocol

MartinHjelmare:
- Here's my take on library design model.
- Please comment and criticize!
- I've included just enough to show the concept I hope, based on the existing code.
~~- Names are hard. Eg I swapped Data for Gateway, but not sure about that one.~~

## Outline

- The model package represents the pieces that together make up the Z-Wave network, controller and sent data.
- The protocol enum and event handlers handles the actions of each model piece. What should happen when a message or event comes in to a piece in the model.


marcelveldt:
Can we settle on what to do with this ? I'm about to code the foundation in the integration and it's more or less dependent of choices made here, especially the event listeners...

MartinHjelmare:
I don't think my PR will change event listeners or event emission. It's just a matter of how we abstract the internals of the library at the moment.

marcelveldt:
No, but we can add that after. For example an event when the initial state is received would be great.
I'd like to add that but maybe better to first merge this one so we have a solid foundation to work from.

MartinHjelmare:
What could change things is if we introduce an additional model layer, but since we should map against zwave-js model we shouldn't do that.

MartinHjelmare:
I've now rebased on the updated model in master branch and applied this PR suggestion to that. I haven't built out the node and value protocols yet. I'll do that if we want to go ahead with this.

test/model/test_driver.py
```diff
@@ -14,6 +14,7 @@ def test_from_state():
     for msg in ws_msgs[1:]:
         msg = json.loads(msg)
         assert msg["type"] == "event"
-        driver.receive_event(msg["event"])
+        driver_event = driver_pkg.DriverEvent(type=msg["event"], data=msg)
+        driver.receive_event(driver_event)
```
MartinHjelmare, line 18: There seems to be a missing key in the fixture data we use here. We get a `KeyError: 'source'`.

MartinHjelmare: Nevermind. It was just me.

marcelveldt: > I've now rebased on the updated model in master branch and applied this PR suggestion to that. I haven't built out the node and value protocols yet. I'll do that if we want to go ahead with this.

Looks good!. I could really use this merged in for the integration as I can reference the enums etc.

test/model/test_driver.py
```diff
@@ -14,6 +14,9 @@ def test_from_state():
     for msg in ws_msgs[1:]:
         msg = json.loads(msg)
         assert msg["type"] == "event"
-        driver.receive_event(msg["event"])
+        driver_event = driver_pkg.DriverEvent(
```
balloob, line 17: These are not driver events. These are all events stored in the dump, which can be any source.

zwave_js_server/model/controller.py
```diff
@@ -1,6 +1,18 @@
...
+from typing import Dict, List, cast
+
 from ..event import EventBase
-from .node import Node
+from ..protocol import ProtocolType, controller as protocol, get_handler
+from .node import Node, NodeEvent
+
+
+@dataclass
+class ControllerEvent:
```
balloob, line 11: We can't distinguish the events in different classes. They all come in at at the root and are forwarded accordingly. They all follow the same format.

What's the benefit from breaking it up into different classes instead of just having an event class?

zwave_js_server/protocol/__init__.py
```diff
@@ -0,0 +1,25 @@
+"""Provide a protocol for zwave-js-server."""
+from typing import Any, Callable, Protocol
+
+
+class ModelType(Protocol):
+    """Represent a model event type."""
+
+    type: Any
```
balloob, line 8:
```suggestion
    type: str
```

zwave_js_server/protocol/__init__.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+class ProtocolType(Protocol):
+    """Represent a protocol module type."""
+
+    Type: Any
+    Handler: Any
+
+
+def get_handler(protocol: ProtocolType, data: ModelType) -> Callable:
```
balloob, line 19: This could be just on a protocol instance.

zwave_js_server/protocol/controller.py
```diff
@@ -0,0 +1,90 @@
...
+    EXCLUSION_STARTED = "exclusion started"
+    INCLUSION_STOPPED = "inclusion stopped"
+    EXCLUSION_STOPPED = "exclusion stopped"
+    NODE_ADDED = "node added"
+    NODE_REMOVED = "node removed"
+    HEAL_NETWORK_PROGRESS = "heal network progress"
+    HEAL_NETWORK_DONE = "heal network done"
+
+
+class Handler:
```
balloob, line 26: I see what's going on here, but it is this really improving the existing logic? I think it's harder to follow. 

Instead of having a Controller instance that has controller properties and methods, we know move the methods to a second class and have that class mutate the public instance variables of a controller? That kinda treats a controller like a C-like struct: just a holder of data. 

These methods are changing instance variables of the models and `event.data`. It fels weird to decouple that. A protocol mutates data and the owner class is not in the loop?

balloob: I think that we can decide to have the `get_handler` logic be included in the models that receive events so that we can still split out handling of each event into different methods.

MartinHjelmare: I think that works.

zwave_js_server/model/controller.py
MartinHjelmare: We can use a single class. This is a leftover from before we settled on mapping the models 1:1 to Z-Wave JS.

test/model/test_driver.py
MartinHjelmare: I'll make a common Event class.

zwave_js_server/event.py
```diff
@@ -1,21 +1,26 @@
 """Event base classes for Z-Wave JS."""
 
 
+from typing import Callable, Dict, List
+
+
 class EventBase:
```
MartinHjelmare, line 7: Should we rename this to EventModelBase?

zwave_js_server/protocol/__init__.py
MartinHjelmare: I've moved this into the EventBase as a helper.

zwave_js_server/protocol/controller.py
MartinHjelmare: I've moved the handlers to each model class.

MartinHjelmare:
The feedback really cleaned things up.

MartinHjelmare:
Converting to draft so we don't merge by mistake.

zwave_js_server/event.py
```diff
@@ -1,21 +1,42 @@
...
+
+    type: str
+    data: dict = field(default_factory=dict)
 
 
 class EventBase:
-    def __init__(self):
+    """Represent a base for event handling model classes."""
+
+    def __init__(self, protocol: Any) -> None:
```
MartinHjelmare, line 17: I'm not sure how to type the protocol in the best way since we want to be able to accept any model protocol Enum.

zwave_js_server/event.py
```diff
@@ -1,21 +1,42 @@
...
         for listener in self._listeners.get(event_name, []):
             listener(data)
+
+    def _handle_event_protocol(self, event: Event) -> None:
+        """Process an event based on event protocol."""
+        protocol_type = self.protocol(event.type)
+        protocol_handler: Callable = getattr(
+            self, f"handle_{protocol_type.name.lower()}"
+        )
+        protocol_handler(event)
```
balloob, line 42: I think that we can do it without a protocol. We can just check for the existence of the methods.

```suggestion
    handler = getattr(self, f"handle_{event.type.replace(' ', '_')}", None)
    if handler is None:
        LOGGER.debug("Received unknown event")
        return
    handler(event)
```

zwave_js_server/model/driver.py
```diff
@@ -1,20 +1,41 @@
...
-            pass
+        self._handle_event_protocol(event)
+
+        self.emit(event.data["event"], event.data)
+
+    def handle_error(self, event: Event) -> None:
+        """Process a driver error event."""
+
+    def handle_driver_ready(self, event: Event) -> None:
+        """Process a driver ready event."""
```
balloob, line 30: We don't forward these. The API only comes online once the driver is ready.

MartinHjelmare: Is it just this event ("driver ready") that we don't handle? We handle "error" and "all nodes ready"?

balloob: `all nodes ready` we do. Error we don't forward because it's unclear to me when we get it. Is it during initialization or runtime too

marcelveldt:
you should probably merge this before it gets too large and address any issue that may still be there in follow up PR's if needed. Overall it's looking good!

MartinHjelmare:
Just waiting for Paulus approval.

balloob (approved):
Nice! 🎉

PR #11, opened 2021-01-07
Add support for start listening

balloob:


zwave_js_server/client.py
```diff
@@ -176,12 +188,23 @@ async def _handle_connection(self):
...
+                version_msg["serverVersion"],
+                version_msg["driverVersion"],
+            )
             self.state = STATE_CONNECTED
 
             if self._on_connect:
                 await gather_callbacks(self._logger, "on_connect", self._on_connect)
 
+            await client.send_json(
+                {"messageId": uuid.uuid4().hex, "command": "start_listening"}
```
marcelveldt, line 205: you might want to move creation of default id to the send_json method perhaps so it can be omitted ?

marcelveldt: but that means the send json needs to check for messageId key

MartinHjelmare (approved):
Python looks good.

zwave_js_server/client.py
balloob: Will do that in the next pass when adding full support for commands.

PR #12, opened 2021-01-07
Address most lint errors

MartinHjelmare:
- The remaining pylint errors should be solved by the event protocol PR.


marcelveldt (approved):
Looks good!

PR #14, opened 2021-01-08
Fix value id and metadata

marcelveldt:
fix value_id creation (propertyKey is optional)
fix typo in atrribute name

### DIFF
--- a/zwave_js_server/model/value.py
+++ b/zwave_js_server/model/value.py
@@ -9,7 +9,7 @@ if TYPE_CHECKING:
 
 def value_id(node: "Node", event: dict) -> str:
     """Return ID of value."""
-    command_class = event.get("commandClass")
+    command_class = event["commandClass"]
     endpoint = event.get("endpoint")
     property_key_name = event.get("propertyKeyName") or event["property"]
     return f"{node.node_id}-{command_class}-{endpoint}-{property_key_name}"
--- a/zwave_js_server/model/value.py
+++ b/zwave_js_server/model/value.py
@@ -11,7 +11,7 @@ def value_id(node: "Node", event: dict) -> str:
     """Return ID of value."""
     command_class = event.get("commandClass")
     endpoint = event.get("endpoint")
-    property_key_name = event.get("propertyKeyName")
+    property_key_name = event.get("propertyKeyName") or event.get("property")
     return f"{node.node_id}-{command_class}-{endpoint}-{property_key_name}"
 
 
@@ -103,7 +103,7 @@ class Value:
         return self.data.get("endpoint")
 
     @property
-    def property_(self) -> str:
+    def property(self) -> str:
         """Return property."""
         return self.data.get("property")
 
--- a/zwave_js_server/model/value.py
+++ b/zwave_js_server/model/value.py
@@ -11,7 +11,7 @@ def value_id(node: "Node", event: dict) -> str:
     """Return ID of value."""
     command_class = event.get("commandClass")
     endpoint = event.get("endpoint")
-    property_key_name = event.get("propertyKeyName") or event.get("property")
+    property_key_name = event.get("propertyKeyName") or event["property"]
     return f"{node.node_id}-{command_class}-{endpoint}-{property_key_name}"
 
 
--- a/zwave_js_server/model/value.py
+++ b/zwave_js_server/model/value.py
@@ -37,9 +37,9 @@ class ValueDataType(TypedDict, total=False):
 
 def value_id(node: "Node", event_data: ValueDataType) -> str:
     """Return ID of value."""
-    command_class = event["commandClass"]
-    endpoint = event.get("endpoint", "00")
-    property_key_name = event.get("propertyKeyName") or event["property"]
+    command_class = event_data["commandClass"]
+    endpoint = event_data.get("endpoint", "00")
+    property_key_name = event_data.get("propertyKeyName") or event_data["property"]
     return f"{node.node_id}-{command_class}-{endpoint}-{property_key_name}"
 
 
--- a/zwave_js_server/model/value.py
+++ b/zwave_js_server/model/value.py
@@ -103,7 +103,7 @@ class Value:
         return self.data.get("endpoint")
 
     @property
-    def property(self) -> str:
+    def property_(self) -> str:
         """Return property."""
         return self.data.get("property")
 
--- a/zwave_js_server/model/value.py
+++ b/zwave_js_server/model/value.py
@@ -58,6 +58,11 @@ class ValueMetadata:
         return self.data.get("unit")
 
     @property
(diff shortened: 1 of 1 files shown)


marcelveldt:
@balloob just a small PR with some missing stuff.

zwave_js_server/model/value.py
```diff
@@ -11,7 +11,7 @@ def value_id(node: "Node", event: dict) -> str:
     """Return ID of value."""
     command_class = event.get("commandClass")
     endpoint = event.get("endpoint")
-    property_key_name = event.get("propertyKeyName")
+    property_key_name = event.get("propertyKeyName") or event.get("property")
```
balloob, line 14: Can we just always use property? It's important that we stay consistent. 

Do we know why keyName can be None?

balloob (approved):

zwave_js_server/model/value.py
marcelveldt: no we can't as this is for the more complex CC's where one CC can have multiple values, fore example the notification CC

marcelveldt: I'll try to find an example if you want. Ran into it while testing

marcelveldt: Schermafbeelding 2021-01-08 om 10 52 44

marcelveldt: It's a sort of (optional) nested structure

zwave_js_server/model/value.py
```diff
@@ -11,7 +11,7 @@ def value_id(node: "Node", event: dict) -> str:
     """Return ID of value."""
     command_class = event.get("commandClass")
     endpoint = event.get("endpoint")
```
MartinHjelmare, line 13: It's weird that we accept None value below in the f-string. Is that really intended?

marcelveldt: property should always be there

MartinHjelmare: If these are not optional we should use `dict[key]`.

marcelveldt: the value id should be:

node_id + endpoint_id + command_class_id + property + optional propertyKey

zwave_js_server/model/value.py
MartinHjelmare:
```suggestion
    property_key_name = event.get("propertyKeyName") or event["property"]
```

zwave_js_server/model/value.py
MartinHjelmare: Endpoint is optional according to the docs.

marcelveldt: I've also changed commandclass as it is mandatory too

MartinHjelmare: So is it ok to have a string `"None"` in the value id for the endpoint position?

balloob: I think it's fine. We can also do `event.get('endpoint', "")` to just have empty string.

marcelveldt: hmm, cosmetical but maybe we should build the string instead or convert it some other presentation

balloob: It's important the `-` are still there to wrap it.

marcelveldt: Leave it as is ?

MartinHjelmare: At minimum we should document the intention in a docstring or comment.

zwave_js_server/model/value.py
```diff
@@ -9,9 +9,9 @@
 
 def value_id(node: "Node", event: dict) -> str:
     """Return ID of value."""
-    command_class = event.get("commandClass")
-    endpoint = event.get("endpoint")
-    property_key_name = event.get("propertyKeyName")
+    command_class = event["commandClass"]
+    endpoint = event.get("endpoint", "")
```
MartinHjelmare, line 13: Should we be explicit instead and have a default like `"no_endpoint"`?

balloob: I think that's fine as the ID is not supposed to be human readable.

MartinHjelmare: Yeah, but we'll still read this when debugging, right? Having a double dash is easy to glance over or misinterpret, I'm thinking.

marcelveldt: I'm fine with some default value like "00" or "no_endpoint".
As far as I can see nodes always have an endpoint.

marcelveldt: @MartinHjelmare changed default value to 00 and resolved the merge conflict

MartinHjelmare (approved):
Thanks!

zwave_js_server/model/value.py
```diff
@@ -37,10 +37,9 @@ class ValueDataType(TypedDict, total=False):
 
 def value_id(node: "Node", event_data: ValueDataType) -> str:
     """Return ID of value."""
-    command_class = event_data.get("commandClass")
-    endpoint = event_data.get("endpoint")
-    property_key_name = event_data.get("propertyKeyName")
-    # TODO: Should we really include None items in the f-string below?
+    command_class = event["commandClass"]
+    endpoint = event.get("endpoint", "00")
+    property_key_name = event.get("propertyKeyName") or event["property"]
```
MartinHjelmare, line 42:
```suggestion
    command_class = event_data["commandClass"]
    endpoint = event_data.get("endpoint", "00")
    property_key_name = event_data.get("propertyKeyName") or event_data["property"]
```

MartinHjelmare (approved):

zwave_js_server/model/value.py
marcelveldt: hmm that happened with merging I guess

zwave_js_server/model/value.py
```diff
@@ -87,6 +86,10 @@ def unit(self) -> Optional[str]:
         return self.data.get("unit")
 
     @property
+    def states(self) -> Optional[dict]:
+        """Return (optional) states."""
+        return self.data.get("states")
```
MartinHjelmare, line 91: We need to add this item to the TypedDict.

marcelveldt: will you do so or should I ?

MartinHjelmare: I did it now.

MartinHjelmare:
Note that I didn't add all the items in the meta data typed dict in my typing effort. I wasn't completely sure about the items since some items that we expected were also not present in the docs. It would be good to look over all of this at some point.

marcelveldt:
> Note that I didn't add all the items in the meta data typed dict in my typing effort. I wasn't completely sure about the items since some items that we expected were also not present in the docs. It would be good to look over all of this at some point.

That's because values can be CC specific

PR #16, opened 2021-01-08
Add once

balloob:
Add `once` helper to listen to an event exactly once.

### DIFF
--- a/zwave_js_server/event.py
+++ b/zwave_js_server/event.py
@@ -1,7 +1,7 @@
 """Provide Event base classes for Z-Wave JS."""
 import logging
 from dataclasses import dataclass, field
-from typing import Any, Callable, Dict, List
+from typing import Callable, Dict, List
 
 LOGGER = logging.getLogger(__package__)
 
--- a/zwave_js_server/event.py
+++ b/zwave_js_server/event.py
@@ -35,6 +35,18 @@ class EventBase:
 
         return unsubscribe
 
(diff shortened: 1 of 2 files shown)


marcelveldt:
Looks fine to me!

zwave_js_server/event.py
```diff
@@ -35,6 +35,18 @@ def unsubscribe() -> None:
 
         return unsubscribe
 
+    def once(self, event_name: str, callback: Callable) -> Callable:
+        """Listen for an event exactly once."""
+        unsub = None
```
MartinHjelmare, line 40: This isn't needed I think.

zwave_js_server/event.py
```diff
@@ -35,6 +35,18 @@ def unsubscribe() -> None:
...
+
+        def event_listener(event):
```
MartinHjelmare, line 42:
```suggestion
        def event_listener(event: str) -> None:
```

zwave_js_server/event.py
```diff
@@ -1,7 +1,7 @@
 """Provide Event base classes for Z-Wave JS."""
 import logging
 from dataclasses import dataclass, field
-from typing import Callable, Dict, List
+from typing import Any, Callable, Dict, List
```
MartinHjelmare, line 4:
```suggestion
from typing import Callable, Dict, List
```

MartinHjelmare (approved):

PR #17, opened 2021-01-08
Disable mypy on CI

balloob:
MyPy has a lot of issues. Let's disable it so we can actually use CI status to check if new things start failing.


MartinHjelmare:
https://github.com/home-assistant-libs/zwave-js-server-python/pull/18

PR #18, opened 2021-01-08
Add missing typing

MartinHjelmare:
- I've marked some questions I had with `# TODO`. We can follow up these later.


marcelveldt:
nice!

PR #22, opened 2021-01-08
Add set_value command to ZwaveNode

marcelveldt:


zwave_js_server/model/node.py
```diff
@@ -216,6 +220,21 @@ def receive_event(self, event: Event) -> None:
 
         self.emit(event.type, event.data)
 
+    async def set_value(self, val: Union[Value, str], new_value: Any) -> None:
```
MartinHjelmare, line 223: I'd like to clarify what kind of value the first parameter is. Not sure what's best.
```suggestion
    async def set_value(self, value_id: Union[Value, str], new_value: Any) -> None:
```

marcelveldt: value_id clashes with the method value_id

marcelveldt: maybe just value ?

MartinHjelmare: Or we decide to use `value` to mean specifically `Value` instances, but it's hard to keep that consistent I'm fearing.

marcelveldt: Well, I think the better option would be to use the naming of upstream lib,
so Value will become ZWaveValue and Node will become ZWaveNode etc.
So for variable names you can refer to zwave_value instead of just value

MartinHjelmare: Sounds good!

marcelveldt: Or go with your proposal. Use NodeValue class name and node_value as instance name

MartinHjelmare: We can do that in a follow up. I'm happy with this PR as is.

MartinHjelmare (approved):

MartinHjelmare:
Do we need to wait for upstream server PR?

zwave_js_server/model/node.py
marcelveldt: Maybe write it on the todo list ?

marcelveldt:
> Do we need to wait for upstream server PR?

Yeah, it depends on this PR:
https://github.com/zwave-js/zwave-js-server/pull/19

zwave_js_server/model/node.py
MartinHjelmare: Can you make an issue?

MartinHjelmare (approved):

marcelveldt:
upstream is fixed. this one can be merged.

PR #23, opened 2021-01-08
Adjust script to recent class changes

marcelveldt:


zwave_js_server/__main__.py
```diff
@@ -62,9 +62,9 @@ def log_value_updated(event: dict) -> None:
         value = event["value"]
 
         if node.device_config:
-            description = node.device_config["description"]
+            description = node.device_config.description
         else:
-            description = f'{node.device_class["generic"]} (missing device config)'
+            description = f'{node.device_class.generic} (missing device config)'
```
MartinHjelmare, line 67: Should we guard against a None value here on `node.device_class.generic`?

zwave_js_server/__main__.py
```diff
@@ -62,9 +62,9 @@ def log_value_updated(event: dict) -> None:
         value = event["value"]
 
         if node.device_config:
-            description = node.device_config["description"]
+            description = node.device_config.description
```
MartinHjelmare, line 65: This may return `None`. Maybe that's ok.

marcelveldt: yeah, no problem

zwave_js_server/__main__.py
marcelveldt: should always be present

MartinHjelmare: It's typed as Optional in the device class.

MartinHjelmare: I'll fix this tomorrow. Merging as is.

MartinHjelmare (approved):

PR #24, opened 2021-01-08
Fix connection callbacks

### LINKED ISSUE #19: connection state
the connection state is set only after the "disconnected" event is fired, this causes a race condition in the integration to set the availability state when client connection state changes.
the event should be fired when the connection state is already been set

### LINKED ISSUE #20: client connection callbacks should return unsubscribe method
Schermafbeelding 2021-01-08 om 16 00 13

marcelveldt:
This fixes #19 and fixes #20

### DIFF
--- a/zwave_js_server/client.py
+++ b/zwave_js_server/client.py
@@ -181,7 +181,9 @@ class Client:
 
             if self.state == STATE_CONNECTED:
                 # change state to connecting/disconnected
-                self.state = STATE_DISCONNECTED if self.close_requested else STATE_CONNECTING
+                self.state = (
+                    STATE_DISCONNECTED if self.close_requested else STATE_CONNECTING
+                )
                 # notify callbacks about disconnection
                 if self._on_disconnect:
                     await gather_callbacks(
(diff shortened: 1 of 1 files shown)


zwave_js_server/client.py
```diff
@@ -156,15 +179,18 @@ async def connect(self) -> None:
                 # Still adding it here to make sure we can always reconnect
                 self._logger.exception("Unexpected error")
 
-            if self.state == STATE_CONNECTED and self._on_disconnect:
-                await gather_callbacks(
-                    self._logger, "on_disconnect", self._on_disconnect
-                )
+            if self.state == STATE_CONNECTED:
+                # change state to connecting
+                self.state = STATE_CONNECTING
```
MartinHjelmare, line 184: I think we need to check `self.close_requested` here too and if that's true the state should be `STATE_DISCONNECTED`. That will handle the client getting disconnected cleanly by the user. In that case we don't want to say that we're connecting to callbacks.

MartinHjelmare (approved):
Nice!

PR #27, opened 2021-01-09
Make CLI shutdown cleaner

MartinHjelmare:
- When `asyncio.run` is closing down due to Keyboard signal, it will ask all tasks to cancel. So the `client.connect` task will receive a cancel exception before we ask the client to disconnect. This causes us to re-enter the retry connection loop unless we check the `close_requested` attribute.
- The flow is still not optimal after this change. The client connect task is still asked to cancel before we disconnect. Optimally we should handle the cancel error outside the client and ask the client to disconnect.


balloob (approved):
You found it! I was looking at this and couldn't figure it out 👏👏

PR #28, opened 2021-01-10
Fix getvalue to use both property and the propertykeyname

marcelveldt:
Fix getvalue to use both property and the propertykeyname
adjust methodname to get_value_id to make its intention more clear


MartinHjelmare (approved):
Looks good!

PR #31, opened 2021-01-11
Allow getting results

balloob:
Add feature to client to send commands and get the results.

```python
try:
    result = await client.async_send_command({ "command": "start_listening" })
except zwave_js_server.client.FailedCommand as err:
    print("Command failed with", err.error_code)
```

### DIFF
--- a/zwave_js_server/client.py
+++ b/zwave_js_server/client.py
@@ -149,7 +149,7 @@ class Client:
 
     async def async_send_command(self, message: Dict[str, Any]) -> dict:
         """Send a command and get a response."""
-        future = self._loop.create_future()
+        future: asyncio.Future[dict] = self._loop.create_future()
         message_id = message["messageId"] = uuid.uuid4().hex
         self._result_futures[message_id] = future
         await self.async_send_json_message(message)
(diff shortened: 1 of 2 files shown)


zwave_js_server/client.py
```diff
@@ -309,3 +318,19 @@ async def disconnect(self) -> None:
 
         if self._disconnect_event is not None:
             await self._disconnect_event.wait()
+
+    async def _start_listening(self):
```
MartinHjelmare, line 322:
```suggestion
    async def _start_listening(self) -> None:
```

zwave_js_server/client.py
```diff
@@ -146,9 +147,18 @@ def connected(self) -> bool:
         """Return if we're currently connected."""
         return self.state == STATE_CONNECTED
 
-    async def async_send_json_message(
-        self, message: Dict[str, Any], message_id: str = None
-    ) -> None:
+    async def async_send_command(self, message: Dict[str, Any]) -> dict:
+        """Send a command and get a response."""
+        future = self._loop.create_future()
```
MartinHjelmare, line 152: I think we need to type the future. I'm not sure what's the best practice for that. I've been looking around but not found a good answer yet. Maybe use `Awaitable`?

MartinHjelmare: This may be relevant:
https://www.python.org/dev/peps/pep-0585/

Maybe:

```py
future: Awaitable[dict] = loop.create_future()

# or

future: Future[dict] = loop.create_future()
```

balloob: Fixe.d

cgarwood (approved):

PR #36, opened 2021-01-12
Add repr with useful info

MartinHjelmare:
- Add some repr methods to make logs more informative.

### DIFF
--- a/zwave_js_server/client.py
+++ b/zwave_js_server/client.py
@@ -73,7 +73,7 @@ class Client:
 
     def __repr__(self) -> str:
         """Return the representation."""
-        return f"{type(self).__name__}(ws_server_url={self.ws_server_url})"
+        return f"{type(self).__name__}(ws_server_url={self.ws_server_url!r})"
 
     def async_handle_message(self, msg: dict) -> None:
         """Handle incoming message.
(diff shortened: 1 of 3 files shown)


zwave_js_server/client.py
```diff
@@ -71,6 +71,10 @@ def __init__(self, ws_server_url: str, aiohttp_session: ClientSession):
         self._result_futures: Dict[str, asyncio.Future] = {}
         self._loop = asyncio.get_running_loop()
 
+    def __repr__(self) -> str:
+        """Return the representation."""
+        return f"{type(self).__name__}(ws_server_url={self.ws_server_url})"
```
balloob, line 76: Should we wrap it like this `<Client(url)>` ?

zwave_js_server/model/node.py
```diff
@@ -60,6 +60,10 @@ def __init__(self, client: "Client", data: NodeDataType) -> None:
             get_value_id(self, val): Value(self, val) for val in data["values"]
         }
 
+    def __repr__(self) -> str:
+        """Return the representation."""
+        return f"{type(self).__name__}(client={self.client}, node_id={self.node_id})"
```
balloob, line 65: I think including `client` would make it too verbose. I would prefer `<Node(id)>`

zwave_js_server/client.py
MartinHjelmare: I normally try to create a string that can be passed to `eval` to create the instance again. Here we don't include all the parameters, so that wouldn't really work, but I'm still using that format.

balloob (approved):

PR #38, opened 2021-01-12
feat: add missing node commands

Michsior14:


zwave_js_server/model/node.py
```diff
@@ -241,6 +241,52 @@ async def async_set_value(self, val: Union[Value, str], new_value: Any) -> None:
...
+        """Send getDefinedValueIDs command to Node."""
+        data = await self.client.async_send_command(
+            {
+                "command": "node.get_defined_value_ids",
+                "nodeId": self.node_id,
+            }
+        )
+        return [
+            Value(self, cast(ValueDataType, valueId))
+            for valueId in list(data["valueIds"])
```
balloob, line 263: Why first convert it to a list? That's not needed

```suggestion
            for valueId in data["valueIds"]
```

Michsior14: I got some issues from `mypy` before. I will try it without it ;)

Michsior14:
@balloob  @MartinHjelmare Should we merge commands related PRs without tests and provide them separately after the preferred test implementation is there or just wait for it?

balloob:
Command stubbing merged so you can add tests now .

zwave_js_server/model/value.py
```diff
@@ -63,9 +63,9 @@ def readable(self) -> Optional[bool]:
         return self.data.get("readable")
 
     @property
-    def writable(self) -> Optional[bool]:
+    def writeable(self) -> Optional[bool]:
         """Return writeable."""
-        return self.data.get("writable")
+        return self.data.get("writeable")
```
MartinHjelmare, line 68: Have we checked this spelling against real data? The docs are inconsistent about what spelling is the correct one.

https://zwave-js.github.io/node-zwave-js/#/api/valueid?id=valuemetadata

Michsior14: Yep, I've checked our fixtures:
- https://github.com/home-assistant-libs/zwave-js-server-python/blob/master/test/fixtures/multisensor_6_state.json#L77
- https://github.com/home-assistant-libs/zwave-js-server-python/blob/master/test/fixtures/basic_dump.txt#L1
and the source code of zwave-js:
- https://github.com/zwave-js/node-zwave-js/blob/0a7bdb15dd50ecc5aa146c12c20b360320b9e169/packages/core/src/values/Metadata.ts#L43

MartinHjelmare: Great!

MartinHjelmare (approved):
Nice!

PR #39, opened 2021-01-12
feat: add missing controller commands

Michsior14:


zwave_js_server/model/controller.py
```diff
@@ -128,6 +128,130 @@ def supports_timers(self) -> Optional[bool]:
...
+    async def async_remove_failed_node(self, node: Union[Node, int]) -> None:
+        """Send removeFailedNode command to Controller."""
+        # a node may be specified as node_id or the node itself
+        if not isinstance(node, Node):
+            node = self.nodes[node]
+        # the node object needs to be send to the server
+        await self.client.async_send_json_message(
+            {
+                "command": "controller.begin_inclusion",
+                "nodeId": node.node_id,
```
balloob, line 173:
```suggestion
        if isinstance(node, Node):
            node = node.node_id
        # the node object needs to be send to the server
        await self.client.async_send_json_message(
            {
                "command": "controller.begin_inclusion",
                "nodeId": node,
```

zwave_js_server/model/controller.py
```diff
@@ -128,6 +128,130 @@ def supports_timers(self) -> Optional[bool]:
...
+        return bool(data["success"])
+
+    async def async_stop_exclusion(self) -> bool:
+        """Send stopExclusion command to Controller."""
+        data = await self.client.async_send_command(
+            {"command": "controller.stop_exclusion"}
+        )
+        return bool(data["success"])
+
+    async def async_remove_failed_node(self, node: Union[Node, int]) -> None:
```
balloob, line 164: Why would we accept both? Does Z-Wave JS accepts both too ?

Michsior14: You are right, I've just copied it from node's `async_set_value`. Will remove it then ;)

zwave_js_server/model/association.py
```diff
@@ -0,0 +1,76 @@
+"""Provide a model for the association."""
+from typing import Dict, List, Optional, TypedDict
+
+
+class AssociationGroupType(TypedDict):
```
balloob, line 5: I would prefer we use a data class.
```suggestion
from dataclasses import dataclass

@dataclass
class AssociationGroupType(TypedDict):
```

And then when you have a dict, you can do

```python
return AssociationGroupType(
  maxNodes=data['maxNodes'],
  # etc
)
```

When getting data out of the dict, use `data['maxNodes']` for required, and `data.get('profile')` for optional. Also make sure to mark the types as `Optional[int]` etc.

Michsior14: Should we keep the `Type` in class name or just skip it?

Michsior14:
This is still blocked by the changes introduced it #38

zwave_js_server/model/association.py
```diff
@@ -0,0 +1,25 @@
+"""Provide a model for the association."""
+from typing import Dict, List, Optional
+from dataclasses import dataclass
+
+
+@dataclass
+class AssociationGroup:
+    """Represent a association group dict type."""
+
+    maxNodes: int  # pylint: disable=invalid-name
```
MartinHjelmare, line 10: We should use snake_case names for the attributes that we expose as interface to users of the library.

balloob (approved):

PR #41, opened 2021-01-13
Store version on client

balloob:


Michsior14 (approved):

pylintrc
```diff
@@ -22,6 +22,7 @@ disable=
   locally-disabled,
   too-few-public-methods,
   too-many-public-methods,
+  too-many-instance-attributes,
```
MartinHjelmare, line 25: We can adjust `max-attributes` to eg 20 instead or disable this locally for the client. I think it's good to keep this warning globally as an indicator that maybe something should be refactored.

PR #43, opened 2021-01-13
Allow command stubbing in tests

balloob:
This allows command stubbing in tests.

### DIFF
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -1,10 +1,8 @@
 """Provide common pytest fixtures."""
 import json
 from typing import List, Tuple
-from unittest.mock import AsyncMock, patch
 
 import pytest
-from aiohttp import ClientSession, ClientWebSocketResponse
 
 from zwave_js_server.client import STATE_CONNECTED, Client
 from zwave_js_server.model.driver import Driver
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -1,5 +1,6 @@
 """Provide common pytest fixtures."""
 import json
+from typing import List, Tuple
 from unittest.mock import AsyncMock, patch
 
 import pytest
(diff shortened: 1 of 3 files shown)


Michsior14 (approved):

test/conftest.py
```diff
@@ -24,39 +23,38 @@ def multisensor_6_state_fixture():
...
+        super().__init__("ws://test.org", None)
+        self.state = STATE_CONNECTED
+        self.mock_responses: List[Tuple[dict, dict]] = []
+        self.mock_commands = []
 
-@pytest.fixture(name="ws_client")
-def ws_client_fixture():
-    """Mock a websocket client."""
-    return AsyncMock(spec_set=ClientWebSocketResponse)
+    async def async_send_command(self, message):
```
MartinHjelmare, line 36: I don't like that we're changing internal things on the client. We should patch in the periphery instead, ie the transport.

test/conftest.py
```diff
@@ -24,39 +23,38 @@ def multisensor_6_state_fixture():
...
-    """Patch uuid4."""
-    uuid4_hex = "1234"
-    with patch("uuid.uuid4") as uuid4:
-        uuid4.return_value.hex = uuid4_hex
-        yield uuid4_hex
+    async def async_send_json_message(self, message):
+        """Send JSON message."""
+        raise RuntimeError("Method not mocked!")
 
+    def mock_command(self, match_command: dict, response: dict):
```
MartinHjelmare, line 49: I'd create a factory fixture that return a function that does the mocking, setting return values etc.

test/conftest.py
balloob: I disagree. Mocking the transport is only needed when you're going to test the actual client. When we're testing our models, we just need a stand-in client.

test/conftest.py
balloob: Factory fixture? It's not a fixture but a mocked command here.

test/conftest.py
MartinHjelmare: If we change the client we'll need to update the tests. I don't like this. Patching in the periphery makes sure we minimize the dependency of the tests on our implementation.

test/conftest.py
MartinHjelmare: I mean a pytest fixture function that returns a function that we can call to apply the mock command and return value.

test/conftest.py
balloob: We would also need to update the server and the code that calls it. That seems like things that should change together.

MartinHjelmare: If we patch the transport we'll only need to change the tests if we change transport. That's a lot less likely to happen than that we change how we use the transport in the client. So all the tests should be more stable in general that way.

Another benefit is that the mocks are more predictable since we always patch in the same place and the interface for that will look the same, regardless of what part of the library we're testing.

balloob: I think that it's a premature optimization to make the architecture flexible enough in case we are going to change the way we're going to send commands. If you feel like it's a showstopper for getting basic command stubbing in the test suite, feel free to close this PR.

MartinHjelmare: I'll open a PR with my suggestion and we can decide then.

MartinHjelmare:
Here's my PR suggestion with a pytest fixture approach:
https://github.com/home-assistant-libs/zwave-js-server-python/pull/44

PR #44, opened 2021-01-13
Add mock command fixture

MartinHjelmare:


test/conftest.py
```diff
@@ -59,6 +60,41 @@ async def client_fixture(loop, client_session, ws_client, uuid4):
...
+    async def set_response(message):
+        """Check the message and set the mocked response if a command matches."""
+        for match_command, response in mock_responses:
+            if all(message[key] == value for key, value in match_command.items()):
+                ack_commands.append(message)
+                received_message = {
+                    "type": "result",
+                    "messageId": uuid4,
+                    "result": response,
+                    "success": response["success"],
```
Michsior14, line 86: Not every response will have `success`.

Michsior14:
While I don't have any preferred way how to maintain the tests, this seems a bit cleaner to me.

test/conftest.py
MartinHjelmare: The result messages should either have success True or success False, right?

MartinHjelmare: I guess we're not limited to test result messages with this fixture.

MartinHjelmare: I'll fix this.

test/conftest.py
```diff
@@ -59,6 +60,41 @@ async def client_fixture(loop, client_session, ws_client, uuid4):
...
+                }
+                client.async_handle_message(received_message)
```
balloob, line 88: If you really want to mock the transport, you would not call this method directly but instead have this message be set as result for the open `msg = await client.receive()` 😉

balloob: (it's fine like this)

test/conftest.py
balloob: Every command has a result right now. So this is fine.

balloob (approved):

PR #45, opened 2021-01-15
Add lock constants

raman325:
Moving lock constants from https://github.com/home-assistant/core/pull/45175 into the library. I originally did not add https://github.com/home-assistant/core/pull/45175/files#diff-1bed63c2518298f6b68c0d6c8325514a97971b914be9a7d6809a1c1b68a631f5R47-R56 but I think it makes sense here as long as we don't tie it to HA constants.

### DIFF
--- a/zwave_js_server/const.py
+++ b/zwave_js_server/const.py
@@ -144,7 +144,6 @@ class DoorLockMode(IntEnum):
     UNKNOWN = 254
     SECURED = 255
 
-
 # Depending on the Commmand Class being used by the lock, the lock state is
 # different so we need a map to track it
 LOCK_CMD_CLASS_TO_LOCKED_STATE_MAP = {
--- a/zwave_js_server/const.py
+++ b/zwave_js_server/const.py
@@ -128,3 +128,30 @@ class CommandClass(IntEnum):
     ZIP_ND = 88
     ZIP_PORTAL = 97
     ZWAVEPLUS_INFO = 94
+
+class DoorLockMode(IntEnum):
+    """Enum with all (known/used) Z-Wave lock states for CommandClass.DOOR_LOCK."""
+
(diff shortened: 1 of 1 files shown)


zwave_js_server/const.py
```diff
@@ -128,3 +129,48 @@ class CommandClass(IntEnum):
     ZIP_ND = 88
     ZIP_PORTAL = 97
     ZWAVEPLUS_INFO = 94
+
+
+# Lock constants
+STATE_LOCKED = "locked"
```
marcelveldt, line 135: where is this one used ?

zwave_js_server/const.py
```diff
@@ -128,3 +129,48 @@ class CommandClass(IntEnum):
...
+}
+
+# Depending on the Command Class being used by the lock, the locked state property
+# is different so we need a map to track it
+CMD_CLASS_TO_PROPERTY_MAP = {
+    CommandClass.DOOR_LOCK: "targetMode",
+    CommandClass.LOCK: "locked",
+}
+
+STATE_TO_ZWAVE_MAP: Dict[int, Dict[str, Union[int, bool]]] = {
```
MartinHjelmare, line 167: I think this one belongs in Home Assistant. It maps the home assistant state to the Z-Wave command class and value, right?

zwave_js_server/const.py
MartinHjelmare: I think it's a home assistant state, so we shouldn't add it here.

zwave_js_server/const.py
marcelveldt: I was wondering the same...

zwave_js_server/const.py
raman325: I will remove, I was debating whether or not it belonged here

MartinHjelmare (approved):
Thanks!

marcelveldt (approved):
Nice, thanks!

MartinHjelmare (approved):

PR #46, opened 2021-01-15
Add climate constants

raman325:
Similar to #45 since I have also opened https://github.com/home-assistant/core/pull/45177

Note that I did not move https://github.com/home-assistant/core/pull/45177/files#diff-2760c32c24dd7ac020c6a5cf3c42d95953fe65dc6e71a006ce522a638d4f3727R133-R147 because it is so HA specific.

Likely need to rebase after #45 gets merged but assuming both are OK conceptually may as well release a new version with both PRs included

### DIFF
--- a/zwave_js_server/const.py
+++ b/zwave_js_server/const.py
@@ -1,5 +1,6 @@
 """Constants for the Z-Wave JS python library."""
 from enum import IntEnum
+from typing import Dict, List
 
 
 class CommandClass(IntEnum):
@@ -130,6 +131,7 @@ class CommandClass(IntEnum):
     ZWAVEPLUS_INFO = 94
 
 
+#Lock constants
 class DoorLockMode(IntEnum):
     """Enum with all (known/used) Z-Wave lock states for CommandClass.DOOR_LOCK."""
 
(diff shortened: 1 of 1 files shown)


zwave_js_server/const.py
```diff
@@ -128,3 +129,53 @@ class CommandClass(IntEnum):
...
+    HEATING_ECON = 11
+    COOLING_ECON = 12
+    AWAY = 13
+    FULL_POWER = 15
+    MANUFACTURER_SPECIFIC = 31
+
+
+# In Z-Wave the modes and presets are both in ThermostatMode.
+# This list contains thermostatmodes we should consider a mode only
+MODES_LIST = [
```
marcelveldt, line 158: please rename to THERMOSTAT_MODES_LIST

zwave_js_server/const.py
```diff
@@ -128,3 +129,53 @@ class CommandClass(IntEnum):
...
+    ThermostatMode.OFF,
+    ThermostatMode.HEAT,
+    ThermostatMode.COOL,
+    ThermostatMode.AUTO,
+    ThermostatMode.AUTO_CHANGE_OVER,
+]
+
+# https://github.com/zwave-js/node-zwave-js/blob/master/packages/zwave-js/src/lib/commandclass/ThermostatSetpointCC.ts#L53-L66
+MODE_SETPOINT_MAP: Dict[int, List[str]] = {
```
marcelveldt, line 167: this one too. It is specific for thermostat CC so it should be in the constant's name imo.

zwave_js_server/const.py
marcelveldt: or just THERMOSTAT_MODES (a little shorter)

zwave_js_server/const.py
marcelveldt: BTW: Isn't this also mapping HA modes to ZW modes ? In that case it belongs to the integration.
@MartinHjelmare what do you think ?

raman325: IMO it is not - it's mapping enum values (Z-Wave protocol specific) to the values that are needed for ZWave JS (the string versions)

marcelveldt: yeah, you're right.

marcelveldt (approved):
great!

MartinHjelmare (approved):

MartinHjelmare:
Lock PR is merged. Rebase here and we can merge this too.

MartinHjelmare (approved):

MartinHjelmare:
Great!

PR #48, opened 2021-01-16
Add more thermostat constants

raman325:
Martin caught some constants that I failed to move over from https://github.com/home-assistant/core/pull/45177 in https://github.com/home-assistant-libs/zwave-js-server-python/pull/46


MartinHjelmare (approved):
Thanks!

PR #49, opened 2021-01-16
Add missing climate constants

raman325:
Added some additional property name constants - I've also switched to using an Enum for setpoint types so that I can iterate through it to watch the correct values if they are found.

### DIFF
--- a/zwave_js_server/const.py
+++ b/zwave_js_server/const.py
@@ -189,8 +189,8 @@ class ThermostatMode(IntEnum):
     MANUFACTURER_SPECIFIC = 31
 
 
-class ThermostatOperatingMode(IntEnum):
-    """Enum with all (known/used) Z-Wave Thermostat OperatingModes."""
+class ThermostatOperatingState(IntEnum):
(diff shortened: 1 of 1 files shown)


zwave_js_server/const.py
```diff
@@ -205,6 +207,24 @@ class ThermostatOperatingMode(IntEnum):
     THIRD_STAGE_AUX_HEAT = 11
 
 
+class ThermostatSetpointType(Enum):
+    """Enuam with all (known/used) Z-Wave Thermostat Setpoint Types."""
```
MartinHjelmare, line 211:
```suggestion
    """Enum with all (known/used) Z-Wave Thermostat Setpoint Types."""
```

MartinHjelmare (commented):
Please also rename this constant to `ThermostatOperatingState`

https://github.com/home-assistant-libs/zwave-js-server-python/blob/02e7e75ccb8858968d0f3cbff6138fa912ad3eac/zwave_js_server/const.py#L190

Michsior14 (approved):

MartinHjelmare (approved):
Thanks!

PR #51, opened 2021-01-16
Add lock utility functions

raman325:
These functions will be used to support HA services and future WS API calls (if we are able to make it that far 😉 )

Note that ZWave JS does not yet support the clear code `Value` - I will add support once it is available

### DIFF
--- a/test/util/test_lock.py
+++ b/test/util/test_lock.py
@@ -1,0 +1,108 @@
+"""Test lock utility functions."""
+import pytest
+from unittest.mock import patch
+
+from zwave_js_server.const import ATTR_CODE_SLOT, ATTR_IN_USE, ATTR_NAME, ATTR_USERCODE
+from zwave_js_server.exceptions import NotFoundError
+from zwave_js_server.util.lock import (
+    get_code_slots,
+    get_usercode,
+    get_usercodes,
+    set_usercode,
+)
+
+CODE_SLOT_NAME = "User Code ({})"
+
+
+def test_get_code_slots(lock_schlage_be469):
+    """Test get_code_slots utility function."""
+    node = lock_schlage_be469
+    code_slots = get_code_slots(node)
+    for x in range(0, 30):
+        code_slot = x + 1
+        in_use = False
+        if x < 3:
+            in_use = True
+        assert code_slots[x] == {
+            ATTR_CODE_SLOT: code_slot,
+            ATTR_IN_USE: in_use,
+            ATTR_NAME: CODE_SLOT_NAME.format(code_slot),
+        }
+
+
+def test_get_usercode(lock_schlage_be469):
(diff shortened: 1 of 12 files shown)


test/util/test_lock.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+def test_get_code_slots(lock_schlage_be469):
+    """Test get_code_slots utility function."""
+    node = lock_schlage_be469
+    code_slots = get_code_slots(node)
+    for x in range(0, 30):
+        code_slot = x + 1
+        in_use = False
+        if x < 3:
```
MartinHjelmare, line 25: Define a constant list of dicts instead with `code_slot`, `in_use` and `user_code` items for all code slots.

test/util/test_lock.py
```diff
@@ -0,0 +1,109 @@
...
+            ATTR_IN_USE: in_use,
+            ATTR_NAME: CODE_SLOT_NAME.format(code_slot),
+            ATTR_USERCODE: usercode,
+        }
+
+
+async def test_set_usercode(lock_schlage_be469):
+    """Test set_usercode utility function."""
+    node = lock_schlage_be469
+    with patch("zwave_js_server.client.Client.async_send_command") as cmd:
```
MartinHjelmare, line 72: We have a fixtures `client` and `mock_command`. See the node tests for example how to use them.

zwave_js_server/util/lock.py
```diff
@@ -0,0 +1,87 @@
...
+                ATTR_CODE_SLOT: int(value.property_key),  # type: ignore
+                ATTR_NAME: value.metadata.label,
+                ATTR_IN_USE: bool(value.value),
+            }
+            if include_usercode:
+                slot[ATTR_USERCODE] = value.value if value.value else None
+
+            slots.append(slot)
+            code_slot += 1
+        except NotFoundError:
```
MartinHjelmare, line 59: Only wrap the line that can raise in the try... except block.

MartinHjelmare (changes requested):

MartinHjelmare (approved):
Looks good!

MartinHjelmare:
Can be merged when build passes.

zwave_js_server/util/lock.py
```diff
@@ -0,0 +1,88 @@
...
+        raise NotFoundError(f"Code slot {code_slot} not found")
+
+    return value
+
+
+def _get_code_slots(
+    node: Node, include_usercode: bool = False
+) -> List[Dict[str, Optional[Union[int, bool, str]]]]:
+    """Get all code slots on the lock and optionally include usercode."""
+    code_slot: int = 1
```
MartinHjelmare, line 40: I don't think we need to type integers.

MartinHjelmare (approved):

PR #55, opened 2021-01-21
Add clear_usercode utility function

raman325:
It was missing from my previous PR because I was unclear on how to clear a code. The upstream logic is here: https://github.com/zwave-js/node-zwave-js/blob/master/packages/zwave-js/src/lib/commandclass/UserCodeCC.ts#L376-L379

### DIFF
--- a/zwave_js_server/util/lock.py
+++ b/zwave_js_server/util/lock.py
@@ -16,21 +16,21 @@ from ..model.node import Node
 from ..model.value import get_value_id, Value
 
 
-def get_code_slot_value(node: Node, code_slot: int, property: str) -> Value:
+def get_code_slot_value(node: Node, code_slot: int, property_name: str) -> Value:
     """Get a code slot value."""
     value = node.values.get(
         get_value_id(
             node,
             {
                 "commandClass": CommandClass.USER_CODE,
-                "property": property,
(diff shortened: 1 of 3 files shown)


zwave_js_server/util/lock.py
```diff
@@ -52,7 +57,7 @@ def _get_code_slots(
         slot = {
             ATTR_CODE_SLOT: int(value.property_key),  # type: ignore
             ATTR_NAME: value.metadata.label,
-            ATTR_IN_USE: bool(value.value),
+            ATTR_IN_USE: status_value.value == CodeSlotStatus.ENABLED,
```
MartinHjelmare, line 60: Are there no tests for this? We should probably tests this since we fixed it here now.

raman325: There is a test for this -> the fixture I used has three codes set, and both sets of logic worked. My understanding of the implementation on the zwave_js side makes the newer logic *more* correct in that it's like the `value_set` property for OZW

MartinHjelmare: Ok. If we can't make it more strict I guess it's ok.

MartinHjelmare (approved):

PR #57, opened 2021-01-21
Add command to dump the state of a server

balloob:
Adds a function to dump the commands from the server. Gets state and optionally keeps connection open to get events.

```
$ python3 -m zwave_js_server --dump-state ws://localhost:3000 --event-timeout 5
{'type': 'version', 'driverVersion': '6.0.0', 'serverVersion': '1.0.0-alpha.2', 'homeId': 1}
{"type":"result","success":true,"result":{"state":{"controller":{"homeId":1},"nodes":[]}}}
```

### DIFF
--- a/zwave_js_server/__main__.py
+++ b/zwave_js_server/__main__.py
@@ -64,7 +64,7 @@ async def handle_dump_state(
     args: argparse.Namespace, session: aiohttp.ClientSession
 ) -> None:
     """Dump the state of the server."""
-    timeout = None if args.event_timeout is None else int(args.event_timeout)
+    timeout = None if args.event_timeout is None else float(args.event_timeout)
     msgs = await dump_msgs(args.url, session, timeout=timeout)
     for msg in msgs:
         print(msg)
--- a/zwave_js_server/__main__.py
+++ b/zwave_js_server/__main__.py
@@ -21,7 +21,7 @@ def get_arguments() -> argparse.Namespace:
         "--server-version", action="store_true", help="Print the version of the server"
     )
     parser.add_argument(
-        "--dump-state", action="store_true", help="Print the version of the server"
+        "--dump-state", action="store_true", help="Dump the driver state"
     )
     parser.add_argument(
         "--event-timeout",
--- a/zwave_js_server/__main__.py
+++ b/zwave_js_server/__main__.py
@@ -7,6 +7,7 @@ import aiohttp
 
 from .client import Client
 from .version import get_server_version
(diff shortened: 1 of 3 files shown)


cgarwood:
Should the node filter be part of this, or will it be a completely separate command?

balloob:
That's for someone else in a follow up PR 👍

zwave_js_server/__main__.py
```diff
@@ -19,6 +20,13 @@ def get_arguments() -> argparse.Namespace:
     parser.add_argument(
         "--server-version", action="store_true", help="Print the version of the server"
     )
+    parser.add_argument(
+        "--dump-state", action="store_true", help="Print the version of the server"
```
MartinHjelmare, line 24:
```suggestion
        "--dump-state", action="store_true", help="Dump the driver state"
```

zwave_js_server/dump.py
```diff
@@ -0,0 +1,36 @@
+"""Dump helper."""
+import asyncio
+from typing import List
+import aiohttp
```
MartinHjelmare, line 4:
```suggestion

import aiohttp
```

zwave_js_server/dump.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+async def dump_msgs(
+    url: str, session: aiohttp.ClientSession, timeout: float = None
```
MartinHjelmare, line 8:
```suggestion
    url: str, session: aiohttp.ClientSession, timeout: Optional[float] = None
```

zwave_js_server/dump.py
```diff
@@ -0,0 +1,36 @@
...
+    state = await client.receive_json()
+    msgs.append(state)
+
+    if timeout is None:
+        await client.close()
+        return msgs
+
+    current_task = asyncio.current_task()
+    assert current_task is not None
+    asyncio.get_running_loop().call_later(timeout, current_task.cancel)
```
MartinHjelmare, line 27: An alternative would be to create a task that receives events from the client and then sleep for the timeout time before canceling the task. That avoids the loop low level api use.

MartinHjelmare (approved):

zwave_js_server/dump.py
```diff
@@ -0,0 +1,37 @@
+"""Dump helper."""
+import asyncio
+from typing import List
```
MartinHjelmare, line 3:
```suggestion
from typing import List, Optional
```

PR #59, opened 2021-01-23
Dump to wait nodes ready

balloob:
If not all nodes are ready, wait until they are and then restart the command.

Also fixes so that we don't listen for 1 event during timeout but all events. (thanks @MartinHjelmare)

### DIFF
--- a/zwave_js_server/dump.py
+++ b/zwave_js_server/dump.py
@@ -1,6 +1,7 @@
 """Dump helper."""
 import asyncio
 from typing import List, Optional
+
 import aiohttp
 
 
--- a/zwave_js_server/dump.py
+++ b/zwave_js_server/dump.py
@@ -1,12 +1,14 @@
 """Dump helper."""
 import asyncio
 from typing import List, Optional
-
 import aiohttp
 
 
 async def dump_msgs(
-    url: str, session: aiohttp.ClientSession, timeout: Optional[float] = None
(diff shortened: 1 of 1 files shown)


zwave_js_server/dump.py
```diff
@@ -16,8 +18,24 @@ async def dump_msgs(
...
 
     await client.send_json({"command": "start_listening"})
-    state = await client.receive_json()
-    msgs.append(state)
+    msg = await client.receive_json()
+    msgs.append(msg)
+    state = msg["result"]["state"]
+
+    # If it's None, old version of the server, ignore it.
+    if wait_nodes_ready and state["driver"].get("allNodesReady") is False:
```
kpine, line 26: Old version of server raises an error since the key doesn't exist.

```
  File "/src/zwave_js_server/dump.py", line 29, in dump_msgs
    if wait_nodes_ready and state["driver"].get("allNodesReady") is False:
KeyError: 'driver'
```

zwave_js_server/dump.py
```diff
@@ -1,12 +1,14 @@
 """Dump helper."""
 import asyncio
 from typing import List, Optional
-
 import aiohttp
```
MartinHjelmare, line 4:
```suggestion

import aiohttp
```

MartinHjelmare (approved):

PR #60, opened 2021-01-24
Add value notification model

marcelveldt:
adds a model for value notification events

https://zwave-js.github.io/node-zwave-js/#/api/node?id=quotvalue-notificationquot

will be used by the integration to fire HA events for these (stateless) value notifications (e.g. scene support for remotes)

### DIFF
--- a/zwave_js_server/model/node.py
+++ b/zwave_js_server/model/node.py
@@ -344,7 +344,7 @@ class Node(EventBase):
 
     def handle_value_notification(self, event: Event) -> None:
         """Process a node value notification event."""
-        event.data["value"] = ValueNotification.from_event(event)
+        event.data["notification"] = ValueNotification.from_event(event)
 
     def handle_metadata_updated(self, event: Event) -> None:
         """Process a node metadata updated event."""
--- a/zwave_js_server/model/node.py
+++ b/zwave_js_server/model/node.py
@@ -4,7 +4,14 @@ from typing import TYPE_CHECKING, Any, List, Optional, TypedDict, Union, cast
 from ..event import Event, EventBase
 from .device_class import DeviceClass, DeviceClassDataType
 from .device_config import DeviceConfig, DeviceConfigDataType
-from .value import Value, ValueDataType, ValueMetadata, MetaDataType, get_value_id
+from .value import (
+    MetaDataType,
+    Value,
+    ValueDataType,
+    ValueMetadata,
+    ValueNotification,
+    get_value_id,
+)
 
 if TYPE_CHECKING:
     from ..client import Client
@@ -337,6 +344,7 @@ class Node(EventBase):
 
     def handle_value_notification(self, event: Event) -> None:
         """Process a node value notification event."""
+        event.data["value"] = ValueNotification.from_event(event)
 
     def handle_metadata_updated(self, event: Event) -> None:
         """Process a node metadata updated event."""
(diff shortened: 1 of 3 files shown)


marcelveldt:
why does pylint ignore the classmethod?

zwave_js_server/model/value.py
```diff
@@ -170,3 +171,32 @@ def receive_event(self, event: Event) -> None:
...
+    command_class_name: str
+    command_class: int
+    endpoint: int
+    property: str
+    value: Any
+    property_name: str
+    metadata: Optional[ValueMetadata] = None
+
+    @classmethod
+    def from_event(cls, event: Event):  # type: ignore
```
MartinHjelmare, line 193:
```suggestion
    def from_event(cls, event: Event) -> "ValueNotification":
```

zwave_js_server/model/value.py
```diff
@@ -170,3 +171,32 @@ def receive_event(self, event: Event) -> None:
...
+        """Parse event message into ValueNotification."""
+        return cls(
+            command_class_name=event.data["args"]["commandClassName"],
+            command_class=event.data["args"]["commandClass"],
+            endpoint=event.data["args"]["endpoint"],
+            property=event.data["args"]["property"],
+            value=event.data["args"].get("value"),
```
MartinHjelmare, line 200: May the value item be missing?

zwave_js_server/model/value.py
```diff
@@ -170,3 +171,32 @@ def receive_event(self, event: Event) -> None:
...
+    https://zwave-js.github.io/node-zwave-js/#/api/node?id=quotvalue-notificationquot
+    """
+
+    command_class_name: str
+    command_class: int
+    endpoint: int
+    property: str
+    value: Any
+    property_name: str
+    metadata: Optional[ValueMetadata] = None
```
MartinHjelmare, line 190: We're not populating the metadata.

MartinHjelmare (changes requested):

MartinHjelmare:
We can disable the `no-self-use` globally in pylintrc. It's complaining as we now have a function body in the method and we don't use `self`.

zwave_js_server/model/node.py
```diff
@@ -337,6 +344,7 @@ def handle_value_removed(self, event: Event) -> None:
 
     def handle_value_notification(self, event: Event) -> None:
         """Process a node value notification event."""
+        event.data["value"] = ValueNotification.from_event(event)
```
MartinHjelmare, line 347: Should we store this under a different key since it's a value notification and not a value?

zwave_js_server/model/value.py
marcelveldt: I've seen cases where it is

zwave_js_server/model/value.py
marcelveldt: Tried that, mypy was still complaining. Maybe that was just me being blind. I'll try again

zwave_js_server/model/node.py
marcelveldt: Not sure. It is basically the same as a regular value except it's stateless.
Perhaps store it in "notification" key is better ?

MartinHjelmare: Yes, I think we should use a different key since it's a different instance.

marcelveldt: done!

MartinHjelmare (approved):
Nice!

PR #61, opened 2021-01-24
Fix KeyError on Value's metadata

marcelveldt:
metadata is optional and thus can be missing in events.
Only initialize it if it's actually present in the data.

This also fixes the issue of nodes being interviewed when HA already started. The node ready event fails because metadata key is missing on one or more values.

### DIFF
--- a/zwave_js_server/model/value.py
+++ b/zwave_js_server/model/value.py
@@ -117,9 +117,9 @@ class Value:
         return get_value_id(self.node, self.data)
 
     @property
-    def metadata(self) -> ValueMetadata:
+    def metadata(self) -> Optional[ValueMetadata]:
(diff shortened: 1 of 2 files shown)


MartinHjelmare (approved):

MartinHjelmare:
Would be good with some tests.

marcelveldt:
What do you want to test? Incoming value without metadata present?

MartinHjelmare:
Yes.

marcelveldt:
superseded by https://github.com/zwave-js/zwave-js-server/pull/62

the real issue is/was that Metadata was only sent at initial state and never after that.
After that change in the lib, it will always have a value and will never be null/None.

Issue #66, opened 2021-01-25
Handle reconnect to Z-Wave JS server


balloob:
The client holds a `driver` object based on the initial received state. This object has listeners attached from Home Assistant.

When we reconnect, the client now holds a driver object that no longer represents the latest state. We don't deal with this case.

I can think of two options: 
 1. We throw away the `driver `object. Consumers of this lib will just have to re-initialize their stack + attach listeners. In the case of HA we would reload the config entry.
 2. We teach each model to update based on the latest state and add new events to indicate the consumer that they have to re-scan each model for latest data/values.

MartinHjelmare:
1. makes most sense to me since the whole state is new. There's no value then to update models individually.

balloob:
I agree. That way we know for sure everything is initialized correctly. 

For Home Assistant, we should add generic logic that if an entry is removed from the state machine and it is still registered in the entity registry, we write unavailable. That way this will go seamless.

balloob:
Generic logic for HA: https://github.com/home-assistant/core/pull/45528

marcelveldt:
Hmmm, this seems a bit complex when only a short hiccup of the connection occurs. I agree that we need some way to ensure we have the latest state but this feels a bit overly complex and, besides... why are we solving the reconnecting HA and not in the library ?

How I would do it:

If the connection is dropped and reconnected, simply ask the server for the full state again and loop through the values. 
We have the full state stored in the lib so we can very easily detect any changes.

Logic to mark the entity as unavailable when the connection is down is already down.

marcelveldt:
But ofcourse there are two possible scenarios:

1) Connection to the WS server got dropped somehow and is reconnected. Only a short period of time and there are possible no or very few changes to process.
2) The whole WS server got reconnected. As the whole mesh is starting and interviewed there are many changes.
3) The whole WS server reconnects to a different (or factory resetted) stick, huge changes to process.

balloob:
In all situations we already mark the integration as unavailable, restarting the integration wouldn't change anything for the user experience once https://github.com/home-assistant/core/pull/45528 is merged.

MartinHjelmare:
This will be fixed by https://github.com/home-assistant-libs/zwave-js-server-python/pull/78 and connected changes in Home Assistant.

PR #67, opened 2021-01-25
Clean up the client

balloob:
- Make `start_listening` command optional
- Rewrite `get_server_version` command to use our client
- Big JSON results will now we parsed in a thread to avoid blocking event loop. Those server states are BIG.
- Add a warning when we reconnect as we don't handle that case yet. Created issue #66 to track it.
- Version info class moved to model folder

### DIFF
--- a/zwave_js_server/version.py
+++ b/zwave_js_server/version.py
@@ -1,4 +1,6 @@
 """Version helper."""
+from typing import cast
+
 import aiohttp
 
 from .client import Client, VersionInfo
@@ -8,10 +10,10 @@ async def get_server_version(url: str, session: aiohttp.ClientSession) -> Versio
     """Return a server version."""
     client = Client(url, session, start_listening_on_connect=False)
 
-    async def handle_connected():
+    async def handle_connected() -> None:
         await client.disconnect()
 
     client.register_on_connect(handle_connected)
 
     await client.connect()
-    return client.version
+    return cast(VersionInfo, client.version)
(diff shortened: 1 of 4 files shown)


zwave_js_server/version.py
```diff
@@ -1,33 +1,17 @@
...
-    """Return a server version."""
-    client = await session.ws_connect(
-        url,
-    )
-    try:
-        return VersionInfo.from_message(await client.receive_json())
-    finally:
-        await client.close()
+    await client.connect()
+    return client.version
```
MartinHjelmare, line 17:
```suggestion
    return cast(VersionInfo, client.version)
```

zwave_js_server/version.py
```diff
@@ -1,33 +1,17 @@
 """Version helper."""
-from dataclasses import dataclass
-
 import aiohttp
```
MartinHjelmare, line 2:
```suggestion
from typing import cast

import aiohttp
```

zwave_js_server/version.py
```diff
@@ -1,33 +1,17 @@
...
 
-    @classmethod
-    def from_message(cls, msg: dict) -> "VersionInfo":
-        """Create a version info from a version message."""
-        return cls(
-            driver_version=msg["driverVersion"],
-            server_version=msg["serverVersion"],
-            home_id=msg["homeId"],
-        )
+    async def handle_connected():
```
balloob, line 11:
```suggestion
    async def handle_connected() -> None:
```

MartinHjelmare (approved):
Looks good!

PR #69, opened 2021-01-26
Fix node state and values

marcelveldt:
This fixes the unpredicted behaviour of processing values before the node is ready.
Belongs to this change in the server: https://github.com/zwave-js/zwave-js-server/pull/68

- Value events will not be forwarded before the node is ready
- Node ready event includes a full dump of the state
- The dump that is received at initialization (start_listening) will only include node values for nodes that are ready

### DIFF
--- a/zwave_js_server/model/node.py
+++ b/zwave_js_server/model/node.py
@@ -62,10 +62,6 @@ class Node(EventBase):
         """Initialize the node."""
         super().__init__()
         self.client = client
-        self.update_data(data)
-
-    def update_data(self, data: NodeDataType) -> None:
-        """Update data when received."""
         self.data = data
         self.values = {
             get_value_id(self, val): Value(self, val) for val in data["values"]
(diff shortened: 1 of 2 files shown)


zwave_js_server/model/node.py
```diff
@@ -62,6 +62,10 @@ def __init__(self, client: "Client", data: NodeDataType) -> None:
         """Initialize the node."""
         super().__init__()
         self.client = client
+        self.update_data(data)
+
+    def update_data(self, data: NodeDataType) -> None:
```
balloob, line 67: We can't call this twice, all old references to values will be lost because we re-create it all. We should update values where possible I think ?

marcelveldt: I was thinking about that too just now... I'll adjust it.

marcelveldt: changed it, up for another review @balloob ?

balloob (approved):

PR #70, opened 2021-01-26
Implement very basic server version check

marcelveldt:
Feel free to give feedback. Implemented this as a very basic server version check as I can see some debugging issues coming because of mismatched server versions.

This will basically enforce a compatible server version and prevent us from doing all kind of backwards compatible checks.

Possible fix for #5

### DIFF
--- a/zwave_js_server/client.py
+++ b/zwave_js_server/client.py
@@ -7,10 +7,12 @@ import uuid
 from typing import Any, Awaitable, Callable, Dict, List, Optional, cast
 
 from aiohttp import ClientSession, ClientWebSocketResponse, WSMsgType, client_exceptions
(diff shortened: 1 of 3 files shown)


zwave_js_server/client.py
```diff
@@ -39,6 +41,10 @@ class InvalidState(Exception):
     """Exception raised when data gets in invalid state."""
 
 
+class InvalidServerVersion(Exception):
+    """Exception raised when connected to server with incompatible major version."""
```
MartinHjelmare, line 45: The check also takes minor version into account.
```suggestion
    """Exception raised when connected to server with incompatible version."""
```

MartinHjelmare (approved):
Looks good! Just one comment.

PR #71, opened 2021-01-26
Bump version to 0.14.0

marcelveldt:
we might want to merge #70 first because the beta.1 server version is a must-have to get some good beta feedback.


marcelveldt:
Once the version is bumped (and released on pypi) we should bump the version of the lib in the HA integration.

MartinHjelmare:
We usually push a version bump commit directly to master branch, when ready.

PR #73, opened 2021-01-28
Fix handling node state

marcelveldt:
This fixes 2 issues:

- On "node ready" event a full state of the node is dumped including values. The new values were not stored and thus HA discovery was failing when a new node was added to the mesh.
- When a node is in the process of being added, first a partial node state is sent, missing most values. Once the interview progresses/completes these values come in. So basically all attributes for the Node model are optional to account for this behaviour.

Fixes HA issue: https://github.com/home-assistant/core/issues/45631

### DIFF
--- a/zwave_js_server/model/device_class.py
+++ b/zwave_js_server/model/device_class.py
@@ -7,7 +7,7 @@ https://zwave-js.github.io/node-zwave-js/#/api/node?id=deviceclass
 from typing import List, Optional, TypedDict
 
 
-class DeviceClassDataType(TypedDict):
+class DeviceClassDataType(TypedDict, total=False):
     """Represent a device class data dict type."""
 
     basic: str
--- a/zwave_js_server/model/node.py
+++ b/zwave_js_server/model/node.py
@@ -24,25 +24,25 @@ class NodeDataType(TypedDict, total=False):
     name: str
     location: str
     status: int  # 0-4  # required
-    deviceClass: DeviceClassDataType  # required
+    deviceClass: DeviceClassDataType
     zwavePlusVersion: int
     nodeType: int
     roleType: int
-    isListening: bool  # required
-    isFrequentListening: bool  # required
-    isRouting: bool  # required
-    maxBaudRate: int  # required
-    isSecure: bool  # required
-    isBeaming: bool  # required
-    version: int  # required
+    isListening: bool
+    isFrequentListening: bool
+    isRouting: bool
+    maxBaudRate: int
+    isSecure: bool
+    isBeaming: bool
+    version: int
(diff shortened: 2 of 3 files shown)


zwave_js_server/model/device_class.py
```diff
@@ -7,7 +7,7 @@
 from typing import List, Optional, TypedDict
 
 
-class DeviceClassDataType(TypedDict):
+class DeviceClassDataType(TypedDict, total=False):
```
cgarwood, line 10: Where is `total` being used?

marcelveldt: mypy didn't like the empty dict coming in...
if the node is being added this info is added at a later stage

MartinHjelmare: `total` is part of the TypedDict api in standard library.

zwave_js_server/model/node.py
```diff
@@ -104,57 +104,57 @@ def ready(self) -> Optional[bool]:
...
     def device_class(self) -> DeviceClass:
         """Return the device_class."""
-        return DeviceClass(self.data["deviceClass"])
+        return DeviceClass(self.data.get("deviceClass", {}))
 
     @property
-    def is_listening(self) -> bool:
+    def is_listening(self) -> Optional[bool]:
         """Return the is_listening."""
         return self.data["isListening"]
```
MartinHjelmare, line 112: If the item is optional we need to use dict.get.

MartinHjelmare:
It would be good to add tests for these cases mentioned in the description.

zwave_js_server/model/node.py
marcelveldt: Oh I forgot one ;-)

marcelveldt:
> It would be good to add tests for these cases mentioned in the description.

How? The info is completely random. Sometimes values XYZ included, sometimes not.
We have to assume a dict with node_id as bare minimum and more info comes in later.

MartinHjelmare:
We simulate the conditions mentioned above.

marcelveldt:
> We simulate the conditions mentioned above and make sure the library acts appropriately.

OK, again, how ? I have no idea. You're the tests wizard ;-)

marcelveldt:
this is what happens btw:
- node inclusion started
- node added event with limited node object (no values and not all keys there)
- node read event (all/most keys there and values list)

MartinHjelmare:
Yes, exactly. Create a node with some minimum state. Then receive the node ready event with needed data and assert that values are updated ok.

marcelveldt:
test added

MartinHjelmare (approved):
Looks good!

PR #74, opened 2021-01-29
Add poll_value command

marcelveldt:
Add's a method to manually poll a Z-Wave value.
This is a best-effort service, allowing users to setup some automation or something to manually poll a value(ID).
It doesn't have a specific return value as it *might* return a value from the node (in any type) or simply nothing.

The idea is to implement this as a entity service in the HA integration with a big fat warning that polling too often is bad ;-)

### DIFF
--- a/test/model/test_node.py
+++ b/test/model/test_node.py
@@ -78,6 +78,24 @@ async def test_set_value(node, uuid4, mock_command):
         "messageId": uuid4,
     }
 
+async def test_poll_value(node, uuid4, mock_command):
+    """Test poll value."""
+    ack_commands = mock_command(
+        {"command": "node.poll_value", "nodeId": node.node_id},
+        {"result": "something"},
+    )
+    value_id = "52-32-00-currentValue-00"
+    value = node.values[value_id]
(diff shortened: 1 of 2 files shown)


marcelveldt:
Depends on https://github.com/zwave-js/zwave-js-server/pull/70/files

test/model/test_node.py
```diff
@@ -79,6 +79,25 @@ async def test_set_value(node, uuid4, mock_command):
...
 
+async def test_poll_value(node, uuid4, mock_command):
+    """Test poll value."""
+    ack_commands = mock_command(
+        {"command": "node.poll_value", "nodeId": node.node_id},
+        {"result": "something"},
+    )
+    value_id = "52-32-00-currentValue-00"
+    value = node.values[value_id]
+    assert await node.async_poll_value(value_id)
```
MartinHjelmare, line 90: Should we check the return value more specifically? We know what it will be in this case.

marcelveldt: done

MartinHjelmare (approved):
Nice!

PR #78, opened 2021-01-31
Rework client

### LINKED ISSUE #66: Handle reconnect to Z-Wave JS server

### LINKED ISSUE #84: Hang when shutting down while a server is offline
Using the `zwave_js` integration in HA, I have two servers configured. Sometimes one of my servers is offline. In this state, the library will attempt to reconnect. If I shutdown HA in this condition, the shutdown will timeout. If all the servers are online, the HA shutdown is nearly instantaneous.

Same result with 0.14.1 and 0.14.2.

```
2021-01-31 22:52:54 DEBUG (MainThread) [zwave_js_server] Trying to connect
2021-01-31 22:52:54 WARNING (MainThread) [zwave_js_server] Unable to connect: Cannot connect to host rpi1:3000 ssl:default [Connect call failed ('192.168.1.20', 3000)]
2021-01-31 22:52:54 INFO (MainThread) [zwave_js_server] Connection closed
2021-01-31 23:01:47 DEBUG (MainThread) [zwave_js_server] Trying to connect
2021-01-31 23:01:47 WARNING (MainThread) [zwave_js_server] Unable to connect: Cannot connect to host rpi1:3000 ssl:default [Connect call failed ('192.168.1.20', 3000)]
2021-01-31 23:01:47 INFO (MainThread) [zwave_js_server] Connection closed
2021-01-31 23:10:53 DEBUG (MainThread) [zwave_js_server] Trying to connect
2021-01-31 23:10:53 WARNING (MainThread) [zwave_js_server] Unable to connect: Cannot connect to host rpi1:3000 ssl:default [Connect call failed ('192.168.1.20', 3000)]
2021-01-31 23:10:53 INFO (MainThread) [zwave_js_server] Connection closed
2021-01-31 23:19:34 DEBUG (MainThread) [zwave_js_server] Trying to connect
2021-01-31 23:19:34 WARNING (MainThread) [zwave_js_server] Unable to connect: Cannot connect to host rpi1:3000 ssl:default [Connect call failed ('192.168.1.20', 3000)]
2021-01-31 23:19:34 INFO (MainThread) [zwave_js_server] Connection closed
^C2021-01-31 23:21:14 INFO (MainThread) [zwave_js_server] Connection closed
2021-01-31 23:21:14 INFO (MainThread) [homeassistant.components.zwave_js] Disconnected from Zwave JS Server
2021-01-31 23:23:14 WARNING (MainThread) [homeassistant.core] Timed out waiting for shutdown stage 1 to complete, the shutdown will continue
```

MartinHjelmare:
I recommend reviewing with hidden white space changes:
https://github.com/home-assistant-libs/zwave-js-server-python/pull/78/files?diff=unified&w=1

- Remove connection re-try feature from client. The library user will be responsible for re-connecting.
- The main methods for connecting, receiving messages and disconnecting are now `Client.connect`, `Client.listen` and `Client.disconnect`, respectively. The library user is responsible for calling these in the correct order, `connect` (1), `listen` (2), `disconnect` (3).
- `Client.listen` now accepts a `driver_ready` `asyncio.Event` parameter to allow the caller to wait for driver to get ready.
- The client callback methods and events have been removed. The library user is responsible for handling connection state propagation.
- Raise `InvalidState` if the client is connected with existing driver state.
- All library exceptions now inherit from `BaseZwaveJSServerError`.
- Adapt CLI script for new client interface.
- Add setup.py entrypoint to install CLI script.

```sh
(zwave-js-server-python) x$ zwave-js-server-python --help
usage: zwave-js-server-python [-h] [--debug] [--server-version] [--dump-state] [--event-timeout EVENT_TIMEOUT] url

Z-Wave JS Server Python

positional arguments:
  url                   URL of server, ie ws://localhost:3000

optional arguments:
  -h, --help            show this help message and exit
  --debug               Log with debug level
  --server-version      Print the version of the server
  --dump-state          Dump the driver state
  --event-timeout EVENT_TIMEOUT
                        How long to listen for events when dumping state
```

- Add client tests.

fixes #66, fixes #84

### DIFF
--- a/zwave_js_server/__main__.py
+++ b/zwave_js_server/__main__.py
@@ -6,10 +6,9 @@ import logging
 import aiohttp
 
 from .client import Client
-from .version import get_server_version
 from .dump import dump_msgs
+from .version import get_server_version
 
-logging.basicConfig(level=logging.INFO)
 logger = logging.getLogger(__package__)
 
 
@@ -17,6 +16,7 @@ def get_arguments() -> argparse.Namespace:
     """Get parsed passed in arguments."""
 
     parser = argparse.ArgumentParser(description="Z-Wave JS Server Python")
+    parser.add_argument("--debug", action="store_true", help="Log with debug level")
     parser.add_argument(
         "--server-version", action="store_true", help="Print the version of the server"
     )
@@ -41,6 +41,9 @@ def get_arguments() -> argparse.Namespace:
 async def main() -> None:
     """Run main."""
     args = get_arguments()
+    level = logging.DEBUG if args.debug else logging.INFO
+    logging.basicConfig(level=level)
+
     async with aiohttp.ClientSession() as session:
         if args.server_version:
             await print_version(args, session)
@@ -74,31 +77,38 @@ async def handle_dump_state(
 async def connect(args: argparse.Namespace, session: aiohttp.ClientSession) -> None:
     """Connect to the server."""
     client = Client(args.url, session)
-    asyncio.create_task(client.connect())
+    asyncio.create_task(register_controller_listeners(client))
 
-    def log_value_updated(event: dict) -> None:
-        node = event["node"]
-        value = event["value"]
+    async with client:
+        await client.listen()
 
-        if node.device_config:
-            description = node.device_config.description
-        else:
-            description = f"{node.device_class.generic} (missing device config)"
 
-        logger.info(
-            "Node %s %s (%s) changed to %s",
-            description,
-            value.property_name or "",
-            value.value_id,
-            value.value,
-        )
+def log_value_updated(event: dict) -> None:
+    """Log node value changes."""
+    node = event["node"]
+    value = event["value"]
 
+    if node.device_config:
+        description = node.device_config.description
+    else:
+        description = f"{node.device_class.generic} (missing device config)"
+
+    logger.info(
+        "Node %s %s (%s) changed to %s",
+        description,
+        value.property_name or "",
+        value.value_id,
+        value.value,
+    )
+
+
+async def register_controller_listeners(client: Client) -> None:
+    """Register controller listeners."""
     is_initialized = asyncio.Event()
 
     async def driver_initialized() -> None:
         """Handle driver init."""
-        assert client.driver is not None  # type
-
+        assert client.driver
         # Set up listeners on new nodes
         client.driver.controller.on(
             "node added",
@@ -113,16 +123,7 @@ async def connect(args: argparse.Namespace, session: aiohttp.ClientSession) -> N
 
     client.register_on_initialized(driver_initialized)
 
-    try:
-        await is_initialized.wait()
-
-        while True:
-            await asyncio.sleep(0.1)
-
-    except asyncio.CancelledError:
-        logger.info("Close requested")
-        await client.disconnect()
-        raise
+    await is_initialized.wait()
 
 
 if __name__ == "__main__":
--- a/zwave_js_server/__main__.py
+++ b/zwave_js_server/__main__.py
@@ -76,10 +76,19 @@ async def handle_dump_state(
 
 async def connect(args: argparse.Namespace, session: aiohttp.ClientSession) -> None:
     """Connect to the server."""
-    client = Client(args.url, session)
-    asyncio.create_task(register_controller_listeners(client))
+    async with Client(args.url, session) as client:
 
-    async with client:
+        assert client.driver
+        # Set up listeners on new nodes
+        client.driver.controller.on(
+            "node added",
+            lambda event: event["node"].on("value updated", log_value_updated),
+        )
+
+        # Set up listeners on existing nodes
+        for node in client.driver.controller.nodes.values():
+            node.on("value updated", log_value_updated)
+
         await client.listen()
 
 
@@ -100,30 +109,6 @@ def log_value_updated(event: dict) -> None:
         value.value_id,
         value.value,
     )
-
-
-async def register_controller_listeners(client: Client) -> None:
-    """Register controller listeners."""
-    is_initialized = asyncio.Event()
-
-    async def driver_initialized() -> None:
-        """Handle driver init."""
-        assert client.driver
-        # Set up listeners on new nodes
-        client.driver.controller.on(
-            "node added",
-            lambda event: event["node"].on("value updated", log_value_updated),
-        )
-
-        # Set up listeners on existing nodes
-        for node in client.driver.controller.nodes.values():
-            node.on("value updated", log_value_updated)
-
-        is_initialized.set()
-
-    client.register_on_initialized(driver_initialized)
-
-    await is_initialized.wait()
 
 
 def main() -> None:
--- a/zwave_js_server/__main__.py
+++ b/zwave_js_server/__main__.py
@@ -78,18 +78,26 @@ async def connect(args: argparse.Namespace, session: aiohttp.ClientSession) -> N
     """Connect to the server."""
     async with Client(args.url, session) as client:
 
-        assert client.driver
-        # Set up listeners on new nodes
-        client.driver.controller.on(
-            "node added",
-            lambda event: event["node"].on("value updated", log_value_updated),
-        )
+        driver_ready = asyncio.Event()
+        asyncio.create_task(on_driver_ready(client, driver_ready))
 
-        # Set up listeners on existing nodes
-        for node in client.driver.controller.nodes.values():
-            node.on("value updated", log_value_updated)
+        await client.listen(driver_ready)
 
-        await client.listen()
+
+async def on_driver_ready(client: Client, driver_ready: asyncio.Event) -> None:
+    """Act on driver ready."""
+    await driver_ready.wait()
+
+    assert client.driver
+    # Set up listeners on new nodes
+    client.driver.controller.on(
+        "node added",
+        lambda event: event["node"].on("value updated", log_value_updated),
+    )
+
+    # Set up listeners on existing nodes
+    for node in client.driver.controller.nodes.values():
+        node.on("value updated", log_value_updated)
 
 
 def log_value_updated(event: dict) -> None:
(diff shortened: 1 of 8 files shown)


zwave_js_server/client.py
```diff
@@ -122,6 +99,10 @@ def async_handle_message(self, msg: dict) -> None:
 
         if msg["type"] != "event":
             # Can't handle
+            self._logger.debug(
+                "Received message with unknown result type with ID: %s",
```
balloob, line 103: Shouldn't we dump full message. Or at least the type

zwave_js_server/client.py
```diff
@@ -173,7 +154,8 @@ def connected(self) -> bool:
 
     async def async_send_command(self, message: Dict[str, Any]) -> dict:
         """Send a command and get a response."""
-        future: "asyncio.Future[dict]" = self._loop.create_future()
+        loop = asyncio.get_running_loop()
```
balloob, line 157: I don't know the overhead of this compared to storing loop as instance variable

zwave_js_server/client.py
```diff
@@ -199,181 +181,119 @@ async def async_send_json_message(self, message: Dict[str, Any]) -> None:
...
 
-                msg_ = cast(dict, msg)
-                try:
-                    self.async_handle_message(msg_)
+        while not self.client.closed:
+            try:
+                msg = await self.client.receive()
+            except (
+                client_exceptions.WSServerHandshakeError,
```
balloob, line 228: We can't get a handshake error. We already received the version

zwave_js_server/client.py
```diff
@@ -199,181 +181,119 @@ async def async_send_json_message(self, message: Dict[str, Any]) -> None:
...
+                client_exceptions.ClientError,
```
balloob, line 229: What client error do you expect here?

zwave_js_server/client.py
```diff
@@ -199,181 +181,119 @@ async def async_send_json_message(self, message: Dict[str, Any]) -> None:
...
-            await self.client.close()
-        elif self.retry_task is not None:
-            self.retry_task.cancel()
+            await self._close()
+
+        self.state = STATE_DISCONNECTED
 
-        if self._disconnect_event is not None:
-            await self._disconnect_event.wait()
+        if self._on_disconnect:
```
balloob, line 268: We now fire these only when we manually call disconnect. How will unexpected disconnect be notified?

zwave_js_server/client.py
MartinHjelmare: https://github.com/python/cpython/blob/e9d4960d15c5282904cf26e469ce7cee39f634f7/Lib/asyncio/events.py#L672-L701

zwave_js_server/client.py
MartinHjelmare: Yeah, this was just a copy when I split the old method.

zwave_js_server/client.py
MartinHjelmare: The docs doesn't mention any exceptions for the receive method. I think the message types ERROR will represent errors.

So I think we can remove the try... except.

zwave_js_server/client.py
MartinHjelmare: The `listen` method will raise an error in all cases except for `WSMsgType.CLOSED, WSMsgType.CLOSING` messages.

The docs are sparse on the message type topic. Is this ok?

zwave_js_server/client.py
MartinHjelmare: It says it's implemented in c.

zwave_js_server/client.py
MartinHjelmare: Just to be clear, the recommended way to use the library is now with the context manager interface. This will disconnect the client upon error.

I'll update the readme with this.

zwave_js_server/client.py
balloob: `self._loop` would be a dict lookup which should be faster since just resolving `asyncio.get_running_loop` is already a dict lookup :)

zwave_js_server/client.py
balloob: The only thing that can be raised here is CancelledError

zwave_js_server/client.py
balloob: We can block calling `listen` by setting a boolean in `aenter` and checking for that.

zwave_js_server/client.py
MartinHjelmare: Ok, I'll switch to a loop attribute client wide.

zwave_js_server/client.py
balloob: That way we don't have to deal with it.

balloob: But I also think that `listen` cancelling should disconnect the client. So the disconnect method just cancels that task.

zwave_js_server/__main__.py
```diff
@@ -74,31 +77,38 @@ async def handle_dump_state(
...
-
-        logger.info(
-            "Node %s %s (%s) changed to %s",
-            description,
-            value.property_name or "",
-            value.value_id,
-            value.value,
-        )
 
+async def register_controller_listeners(client: Client) -> None:
```
balloob, line 105: This can just be inlined in the connect function. We no longer need to wait for `initialized` event or run it as a new task.

balloob: In fact, we could run into a race condition right now where initialized is achieved before the listener is attached.

zwave_js_server/client.py
```diff
@@ -199,181 +121,122 @@ async def async_send_json_message(self, message: Dict[str, Any]) -> None:
...
 
-        finally:
-            if disconnect_warn is None:
-                self._logger.info("Connection closed")
-            else:
-                self._logger.warning("Connection closed: %s", disconnect_warn)
+            try:
+                self.async_handle_message(msg_)
+            except InvalidState as err:
+                await self._close(err)
```
MartinHjelmare, line 219: I wonder if we shoult not close here, but just let the exception be raised? We handle the exception during `connect` now, and if the connect hasn't finished there's no point in closing. It's also the repsonsibility of the library user now to disconnect.

MartinHjelmare: A reason to still close here is to cancel the futures waiting for a received message.

MartinHjelmare: We probably need to close the connection then for all exception raised in listen.

MartinHjelmare: I've simplified the error propagation between listen and connect now.

zwave_js_server/client.py
```diff
@@ -199,181 +121,118 @@ async def async_send_json_message(self, message: Dict[str, Any]) -> None:
...
-                if msg.type in (WSMsgType.CLOSED, WSMsgType.CLOSING):
-                    break
+        listen_task = asyncio.create_task(self.listen())
+        start_listen_task = asyncio.create_task(self._start_listen())
 
-                if msg.type == WSMsgType.ERROR:
-                    disconnect_warn = "Connection error"
-                    break
+        try:
+            await asyncio.gather(listen_task, start_listen_task)
```
MartinHjelmare, line 160: I need to stop of first result here.

zwave_js_server/client.py
```diff
@@ -2,95 +2,51 @@
...
-        self._disconnect_event: Optional[asyncio.Event] = None
-        self._result_futures: Dict[str, asyncio.Future] = {}
         self._loop = asyncio.get_running_loop()
-
-        if start_listening_on_connect:
-            self.register_on_connect(self._start_listening)
+        self._result_futures: Dict[str, asyncio.Future] = {}
 
     def __repr__(self) -> str:
         """Return the representation."""
```
balloob, line 52: Should `async_handle_message` be `_async_handle_message` ?

MartinHjelmare: Yeah, probably. I also think we should remove the `async_` prefix from the client methods.

zwave_js_server/client.py
```diff
@@ -386,3 +247,14 @@ def _check_server_version(self, server_version: str) -> None:
                 "Connected to a Zwave JS Server with an untested version, \
```
balloob, line 247: Should we really warn for this? We're going to reach this very quick. I don't think we should.

zwave_js_server/client.py
```diff
@@ -199,181 +121,120 @@ async def async_send_json_message(self, message: Dict[str, Any]) -> None:
...
-                )
-
-            loop = asyncio.get_running_loop()
-
-            while not client.closed:
-                msg = await client.receive()
-
-                if msg.type in (WSMsgType.CLOSED, WSMsgType.CLOSING):
-                    break
+        listen_task = asyncio.create_task(self.listen())
```
balloob, line 156: In your description you mention the consumer has to call `listen` but here you call it inside `connect`. This will make sure `connect` never finishes.

balloob: That's not good, as we want to wrap connecting + receiving the version message in a timeout.

zwave_js_server/client.py
MartinHjelmare: Not sure. We can think about it and address later. :+1:

marcelveldt: Maybe only when the minor version changes ? Not when the patch version is different.

marcelveldt: Or simply do not warn. We have the minimal server version constant if we have breaking api changes

balloob (approved):

zwave_js_server/client.py
balloob: We should only see if major version is higher and then warn. Rest should be safe if we follow semver

PR #80, opened 2021-01-31
Add endpoint model and endpoints property to node

raman325:
The library is missing the endpoints property in the node model. This adds support.

This came from a discussion with @marcelveldt about an issue I am having with my thermostat where some of the values live on a different endpoint from the primary value. The plan is to enhance `get_zwave_value` (https://github.com/home-assistant/core/blob/dev/homeassistant/components/zwave_js/entity.py#L117) to optionally search additional endpoints if the value can't be found on the primary value's endpoint.


zwave_js_server/model/endpoint.py
```diff
@@ -0,0 +1,44 @@
...
+
+from typing import Optional, TypedDict
+
+
+class EndpointDataType(TypedDict, total=False):
+    """Represent an endpoint data dict type."""
+
+    nodeId: int
+    index: int
+    installerIcon: Optional[int]
```
MartinHjelmare, line 15: Mark the others required with a comment instead. I'm assuming the optional items will not be present at all in the data dict if unset. `Optional` means they can be `None` which may not be the case.

zwave_js_server/model/node.py
```diff
@@ -201,6 +203,11 @@ def neighbors(self) -> List[int]:
         """Return the neighbors."""
         return self.data.get("neighbors", [])
 
+    @property
+    def endpoints(self) -> Optional[List[Endpoint]]:
```
MartinHjelmare, line 207:
```suggestion
    def endpoints(self) -> List[Endpoint]:
```

MartinHjelmare (changes requested):

MartinHjelmare:
We should represent the data as modeled by Z-Wave JS. So if it duplicates the node id we should also do that.

zwave_js_server/model/notification.py
```diff
@@ -0,0 +1,37 @@
+"""
+Model for a Zwave Node's Notification Event.
```
MartinHjelmare, line 2: This seems unrelated to the PR.

zwave_js_server/model/endpoint.py
raman325: yes that's correct

zwave_js_server/model/node.py
```diff
@@ -201,6 +203,11 @@ def neighbors(self) -> List[int]:
         """Return the neighbors."""
         return self.data.get("neighbors", [])
 
+    @property
+    def endpoints(self) -> List[Endpoint]:
+        """Return the endpoints."""
+        return [Endpoint(endpoint) for endpoint in self.data["endpoints"]]
```
MartinHjelmare, line 209: Will the `endpoints` item always be present in the node data?

raman325: I believe this means that it does: https://github.com/zwave-js/zwave-js-server/blob/master/src/lib/state.ts#L98

I don't really know TypeScript but I am assuming that worst case scenario, it will be an empty array

MartinHjelmare: I agree:
https://zwave-js.github.io/node-zwave-js/#/api/node?id=getallendpoints

MartinHjelmare (approved):
Nice!

Issue #25, opened 2021-01-08
consistent class and instance names


marcelveldt:
We already had a few cases we the name "value" was confusing.

Proposal:

rename Node class to ZwaveNode and when referring to an instance of it, use zwave_node

rename Value class to ZwaveValue and when referring to an instance of it, use zwave_value

PR #85, opened 2021-02-01
Add Notification model and handle notification event

raman325:
I plan to submit a PR similar to https://github.com/home-assistant/core/pull/45729 where we fire an event when a notification is received for a node.

### DIFF
--- a/zwave_js_server/model/node.py
+++ b/zwave_js_server/model/node.py
@@ -367,7 +367,7 @@ class Node(EventBase):
 
     def handle_value_notification(self, event: Event) -> None:
         """Process a node value notification event."""
-        event.data["notification"] = ValueNotification(self, event.data["args"])
+        event.data["value_notification"] = ValueNotification(self, event.data["args"])
 
     def handle_metadata_updated(self, event: Event) -> None:
         """Process a node metadata updated event."""
--- a/zwave_js_server/model/node.py
+++ b/zwave_js_server/model/node.py
@@ -1,5 +1,6 @@
 """Provide a model for the Z-Wave JS node."""
 from typing import TYPE_CHECKING, Any, List, Optional, TypedDict, Union, cast
+from zwave_js_server.model.notification import Notification
 
 from ..event import Event, EventBase
 from .device_class import DeviceClass, DeviceClassDataType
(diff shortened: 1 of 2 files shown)


zwave_js_server/model/node.py
```diff
@@ -366,6 +367,9 @@ def handle_metadata_updated(self, event: Event) -> None:
 
     def handle_notification(self, event: Event) -> None:
         """Process a node notification event."""
+        event.data["notification"] = Notification(
```
MartinHjelmare, line 370: I think we should change the key above when storing the `ValueNotification` to `"value_notification"` to be consistent.

raman325: @marcelveldt ^ guess I shouldn't have crossed out my comment on the other PR 🙂 

Do we do that as part of this PR or a separate one?

MartinHjelmare: I think it's ok to do it here. We can bump and release after this is merged.

MartinHjelmare:
Please rebase to solve the merge conflict.

zwave_js_server/model/node.py
marcelveldt: Go ahead and change here. I'll adjust my PR

marcelveldt: BTW: I more and more vote for changing the models with prefixes such as ZwaveNode and ZwaveValue and NodeNotification and ValueNotification, see #25

marcelveldt:
@raman325 maybe use the same logic to fire the event ? I can adjust my PR so it will also forward these events ?

raman325:
The event data will be different but otherwise yeah that makes sense. I was thinking it would be nice eventually to tie these events to the primary "entity" but that might be easier said than done.

zwave_js_server/model/node.py
MartinHjelmare: Yes, we can do that clean up after this PR if merged.

MartinHjelmare:
Entities don't fire events, besides the built in state changed events, in Home Assistant.

MartinHjelmare (approved):
Are we happy to merge?

raman325:
sure I meant add entity_id as part of the event data to use as criteria for automations

MartinHjelmare:
We can discuss that.

PR #89, opened 2021-02-02
Handle metadata updated

marcelveldt:
Belongs to https://github.com/zwave-js/zwave-js-server/pull/81
Fixes missing value.value for (scene activation) events.

This fix should be merged before HA release 2021.2

### DIFF
--- a/zwave_js_server/model/node.py
+++ b/zwave_js_server/model/node.py
@@ -367,10 +367,16 @@ class Node(EventBase):
 
     def handle_value_notification(self, event: Event) -> None:
         """Process a node value notification event."""
+        # append metadata if value metadata is available
+        value = self.values.get(get_value_id(self, event.data["args"]))
+        if value and value.data.get("metadata"):
+            event.data["args"]["metadata"] = value.data["metadata"]
         event.data["value_notification"] = ValueNotification(self, event.data["args"])
 
     def handle_metadata_updated(self, event: Event) -> None:
         """Process a node metadata updated event."""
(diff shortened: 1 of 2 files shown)


zwave_js_server/model/value.py
```diff
@@ -181,7 +181,10 @@ def receive_event(self, event: Event) -> None:
     def update(self, data: ValueDataType) -> None:
         """Update data."""
         self.data.update(data)
-        self._value = data.get("newValue", data.get("value"))
+        if "newValue" in data:
```
balloob, line 184: This is fine, but better would be in a future PR if we check event type so we know how to process the data.

balloob (approved):

zwave_js_server/model/node.py
```diff
@@ -367,10 +367,16 @@ def handle_value_removed(self, event: Event) -> None:
 
     def handle_value_notification(self, event: Event) -> None:
         """Process a node value notification event."""
+        # append metadata if value metadata is available
+        value = self.values.get(get_value_id(self, event.data["args"]))
+        if value and value.data.get("metadata"):
+            event.data["args"]["metadata"] = value.data["metadata"]
```
balloob, line 373: Fine here too, but also something in a future PR where we could maybe forward processing to the value so it can process this event properly.

zwave_js_server/model/value.py
marcelveldt: yes agreed.

PR #90, opened 2021-02-02
Metadata can be optional

marcelveldt:
Due to a small bug in Z-Wave JS the metadata can be missing for a CommandClass Value.
This fix will at least prevent the KeyError when accessing metadata of a value. I've decided to initialize it with an (almost) empty dict to keep compatibility.

We can remove it once the upstream issue is resolved.

https://github.com/zwave-js/node-zwave-js/issues/1560


MartinHjelmare (approved):
Looks good!

PR #92, opened 2021-02-02
Make send JSON message internal

balloob:
This makes send JSON message command on the client internal. It found a bunch of places where we weren't using the send command version and so didn't wait until the command was considered done by the driver.

### DIFF
--- a/zwave_js_server/model/controller.py
+++ b/zwave_js_server/model/controller.py
@@ -167,7 +167,7 @@ class Controller(EventBase):
 
     async def async_remove_failed_node(self, node_id: int) -> None:
         """Send removeFailedNode command to Controller."""
-        await self.client.async_send_json_message(
+        await self.client.async_send_command(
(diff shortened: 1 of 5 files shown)


zwave_js_server/model/controller.py
```diff
@@ -167,7 +167,7 @@ async def async_stop_exclusion(self) -> bool:
 
     async def async_remove_failed_node(self, node_id: int) -> None:
         """Send removeFailedNode command to Controller."""
-        await self.client.async_send_json_message(
+        await self.client.async_send_command(
```
MartinHjelmare, line 170: Should we rename `async_send_command` to just `send_command`?

balloob: Yes, but let's do that another time. I didn't want to make a breaking change in this PR.

MartinHjelmare (approved):

PR #93, opened 2021-02-03
Clean up client connection

balloob:
This cleans up the `listen` command. When the listen phase ends for whatever reason (error, server closing down, disconnect requested), we clean it up right away as we don't allow the object to be used anymore. This way we're sure.

I got stuck with this PR on the tests. There are too many fixtures and it looks like everything is happening somewhere else. So putting this up as draft.

### DIFF
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -1,6 +1,7 @@
 """Provide common pytest fixtures."""
 import asyncio
 import json
+import logging
 from typing import List, Tuple
 from unittest.mock import AsyncMock, Mock, patch
 
@@ -15,6 +16,8 @@ from zwave_js_server.model.driver import Driver
 from zwave_js_server.model.node import Node
 
 from . import load_fixture
+
+logging.basicConfig(level=logging.DEBUG)
 
 # pylint: disable=unused-argument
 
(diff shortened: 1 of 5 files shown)


test/conftest.py
```diff
@@ -16,6 +17,8 @@
 
 from . import load_fixture
 
+logging.basicConfig(level=logging.DEBUG)
```
MartinHjelmare, line 20: This can also be changed via pytest CLI eg:
```
pytest --log-level=DEBUG
```
Not sure if we want to default to debug. It spams a lot.

balloob: It's the kind of spam you want to see if a test breaks

MartinHjelmare: Yes, but sometimes it's in the way and too much scrolling.

balloob: ok, will remove.

zwave_js_server/client.py
```diff
@@ -40,7 +37,6 @@ def __init__(self, ws_server_url: str, aiohttp_session: ClientSession):
         # The WebSocket client
         self._client: Optional[ClientWebSocketResponse] = None
         # Current state of the connection
-        self.state = STATE_DISCONNECTED
```
MartinHjelmare, line 44: This is a breaking change. We need to remember to update the websocket api in the core for this.

balloob: We should be able to easily migrate to using `connected`

MartinHjelmare (approved):

PR #98, opened 2021-02-04
Improve version check

balloob:
- Set a max version that we won't support (2.0)
- Fail if server is max version or higher

### DIFF
--- a/zwave_js_server/client.py
+++ b/zwave_js_server/client.py
@@ -7,7 +7,7 @@ from types import TracebackType
 from typing import Any, Dict, Optional, cast
 
 from aiohttp import ClientSession, ClientWebSocketResponse, WSMsgType, client_exceptions
-from awesomeversion import AwesomeVersion, AwesomeVersionException
+from awesomeversion import AwesomeVersion
 from awesomeversion.strategy import AwesomeVersionStrategy
 
 from .const import MAX_SERVER_VERSION, MIN_SERVER_VERSION
--- a/zwave_js_server/client.py
+++ b/zwave_js_server/client.py
@@ -91,11 +91,20 @@ class Client:
         bad_version = None
 
         if cur_version.strategy != AwesomeVersionStrategy.SEMVER:
-            bad_version = f"Failed to parse Z-Wave JS Server version {self.version.server_version}"
+            bad_version = (
+                "Failed to parse Z-Wave JS Server version "
+                f"{self.version.server_version}"
+            )
         elif cur_version < AwesomeVersion(MIN_SERVER_VERSION):
-            bad_version = f"Z-Wave JS Server needs to be at least version {MIN_SERVER_VERSION}. Found {self.version.server_version}"
+            bad_version = (
+                f"Z-Wave JS Server needs to be at least version {MIN_SERVER_VERSION}. "
+                f"Found {self.version.server_version}"
+            )
         elif cur_version >= AwesomeVersion(MAX_SERVER_VERSION):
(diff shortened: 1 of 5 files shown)


zwave_js_server/const.py
```diff
@@ -3,6 +3,8 @@
 from typing import Dict, List
 
 MIN_SERVER_VERSION = "1.0.0-beta.2"
```
MartinHjelmare, line 5: Is this min version still ok? The add-on runs beta 4 now.

zwave_js_server/client.py
```diff
@@ -86,16 +87,19 @@ async def connect(self) -> None:
...
-        if cur_version < min_version:
+        cur_version = AwesomeVersion(self.version.server_version)
+        bad_version = None
+
+        if cur_version.strategy != AwesomeVersionStrategy.SEMVER:
+            bad_version = f"Failed to parse Z-Wave JS Server version {self.version.server_version}"
+        elif cur_version < AwesomeVersion(MIN_SERVER_VERSION):
+            bad_version = f"Z-Wave JS Server needs to be at least version {MIN_SERVER_VERSION}. Found {self.version.server_version}"
+        elif cur_version >= AwesomeVersion(MAX_SERVER_VERSION):
+            bad_version = f"Z-Wave JS Server is newer or equal to {MAX_SERVER_VERSION}. Found {self.version.server_version}"
```
MartinHjelmare, line 98:
```suggestion
            bad_version = (
                "Failed to parse Z-Wave JS Server version "
                f"{self.version.server_version}"
            )
        elif cur_version < AwesomeVersion(MIN_SERVER_VERSION):
            bad_version = (
                f"Z-Wave JS Server needs to be at least version {MIN_SERVER_VERSION}. "
                f"Found {self.version.server_version}"
            )
        elif cur_version >= AwesomeVersion(MAX_SERVER_VERSION):
            bad_version = (
                f"Z-Wave JS Server is newer or equal to {MAX_SERVER_VERSION}. "
                f"Found {self.version.server_version}"
            )
```

zwave_js_server/client.py
```diff
@@ -7,9 +7,10 @@
 from typing import Any, Dict, Optional, cast
 
 from aiohttp import ClientSession, ClientWebSocketResponse, WSMsgType, client_exceptions
-from packaging.version import parse as parse_version
+from awesomeversion import AwesomeVersion, AwesomeVersionException
```
MartinHjelmare, line 10: The exception is never raised in awesomeversion.
```suggestion
from awesomeversion import AwesomeVersion
```

zwave_js_server/const.py
balloob: I don't know. We are compatible with beta 2+ right ? It's not about the bugs, it's about the protocol.

zwave_js_server/client.py
balloob: Yeah, it's coming in the next version that is being worked on .

zwave_js_server/const.py
marcelveldt: beta4 solved the missing keys in the state dumps so maybe require that as minimum

zwave_js_server/client.py
balloob: Next version is hitting a delay. Just going to remove this :)

MartinHjelmare (approved):

zwave_js_server/const.py
MartinHjelmare: Let's bump this separately.

PR #101, opened 2021-02-05
Add dump tests

MartinHjelmare:


test/conftest.py
```diff
@@ -16,7 +16,7 @@
 
 from . import load_fixture
 
-# pylint: disable=unused-argument
+# pylint: disable=protected-access, unused-argument
```
balloob, line 19: We should not pylint our tests because it's a hell

balloob (approved):

test/conftest.py
MartinHjelmare: We don't do it on CI.

PR #105, opened 2021-02-05
Add configuration value tools and guard Node.async_set_value

raman325:
A couple of changes in this PR:

- There's a new `Value` subclass called `ConfigurationValue` which provides additional metadata about a configuration value.
- Added node helper methods to get values for a specific command class and to get configuration values
- Added guard logic to `async_set_value` for values that aren't writeable and for configuration values that are unbound in the metadata. Per AlCalzone, for unbound configuration values, we need to use the Configuration CC API directly instead of `set_value` which the library doesn't support yet

I'm not sure if this is the best architectural approach to handle configuration values and am open to input there

### DIFF
--- a/zwave_js_server/model/node.py
+++ b/zwave_js_server/model/node.py
@@ -1,5 +1,6 @@
 """Provide a model for the Z-Wave JS node."""
-from typing import TYPE_CHECKING, Any, List, Optional, TypedDict, Union, cast
+from typing import TYPE_CHECKING, Any, Dict, List, Optional, TypedDict, Union, cast
+from zwave_js_server.const import CommandClass, ConfigurationValueType
 
 from ..event import Event, EventBase
 from .device_class import DeviceClass, DeviceClassDataType
@@ -7,6 +8,7 @@ from .device_config import DeviceConfig, DeviceConfigDataType
 from .endpoint import Endpoint, EndpointDataType
 from .notification import Notification, NotificationDataType
 from .value import (
+    ConfigurationValue,
     MetaDataType,
     Value,
     ValueDataType,
@@ -66,9 +68,13 @@ class Node(EventBase):
         super().__init__()
         self.client = client
         self.data = data
-        self.values = {
-            get_value_id(self, val): Value(self, val) for val in data["values"]
-        }
+        self.values = {}
+        for val in data["values"]:
+            value_id = get_value_id(self, val)
+            if val["commandClass"] == CommandClass.CONFIGURATION:
+                self.values[value_id] = ConfigurationValue(self, val)
+            else:
+                self.values[value_id] = Value(self, val)
 
     def __repr__(self) -> str:
         """Return the representation."""
@@ -234,6 +240,25 @@ class Node(EventBase):
         """Return the interview_attempts."""
         return self.data.get("interviewAttempts")
 
+    def get_command_class_values(
+        self, command_class: CommandClass, endpoint: int = None
+    ) -> Optional[Dict[str, Value]]:
+        """Return all values for a given command class."""
+        return {
+            value_id: value
+            for value_id, value in self.values.items()
+            if value.command_class == command_class
+            and (endpoint is None or value.endpoint == endpoint)
+        }
+
+    def get_configuration_values(
+        self, endpoint: int = None
+    ) -> Optional[Dict[str, ConfigurationValue]]:
+        """Return all configuration values for a node."""
+        return self.get_command_class_values(
+            CommandClass.CONFIGURATION, endpoint=endpoint
+        )
+
     def receive_event(self, event: Event) -> None:
         """Receive an event."""
         self._handle_event_protocol(event)
@@ -246,6 +271,19 @@ class Node(EventBase):
         # a value may be specified as value_id or the value itself
         if not isinstance(val, Value):
             val = self.values[val]
+
+        if not val.metadata.writeable:
+            raise TypeError("This configuration value is read only")
+
+        if (
+            isinstance(val, ConfigurationValue)
+            and val.type == ConfigurationValueType.UNDEFINED
+        ):
+            # We need to use the Configuration CC API to set the value for this type
+            raise NotImplementedError(
+                "Configuration values of undefined type can't be set"
+            )
+
         # the value object needs to be send to the server
         result = await self.client.async_send_command(
             {
--- a/zwave_js_server/model/node.py
+++ b/zwave_js_server/model/node.py
@@ -288,19 +288,19 @@ class Node(EventBase):
                     "Configuration values of undefined type can't be set"
                 )
 
-            max = val.metadata.max
-            min = val.metadata.min
+            max_ = val.metadata.max
+            min_ = val.metadata.min
             if val.type == ConfigurationValueType.RANGE and (
-                (max is not None and new_value > max)
-                or (min is not None and new_value < min)
+                (max_ is not None and new_value > max_)
+                or (min_ is not None and new_value < min_)
             ):
                 bounds = []
-                if min is not None:
-                    bounds.append(f"Min: {min}")
-                if max is not None:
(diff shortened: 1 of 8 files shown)


zwave_js_server/model/node.py
```diff
@@ -234,6 +240,25 @@ def interview_attempts(self) -> Optional[int]:
         """Return the interview_attempts."""
         return self.data.get("interviewAttempts")
 
+    def get_command_class_values(
+        self, command_class: CommandClass, endpoint: int = None
+    ) -> Optional[Dict[str, Value]]:
```
MartinHjelmare, line 245: We always return a dict.
```suggestion
    ) -> Dict[str, Value]:
```

zwave_js_server/model/node.py
```diff
@@ -234,6 +240,25 @@ def interview_attempts(self) -> Optional[int]:
...
+        return {
+            value_id: value
+            for value_id, value in self.values.items()
+            if value.command_class == command_class
+            and (endpoint is None or value.endpoint == endpoint)
+        }
+
+    def get_configuration_values(
+        self, endpoint: int = None
+    ) -> Optional[Dict[str, ConfigurationValue]]:
```
MartinHjelmare, line 256:
```suggestion
    ) -> Dict[str, ConfigurationValue]:
```

zwave_js_server/model/node.py
```diff
@@ -246,6 +271,19 @@ async def async_set_value(self, val: Union[Value, str], new_value: Any) -> bool:
         # a value may be specified as value_id or the value itself
         if not isinstance(val, Value):
             val = self.values[val]
+
+        if not val.metadata.writeable:
+            raise TypeError("This configuration value is read only")
```
MartinHjelmare, line 276: If the exception is library user facing the exception should inherit from the library base exception.

If it's just for library developers, we don't need to be as stringent.

raman325: I'm not sure at this point, so safer to inherit from the library base exception

raman325:
any thoughts on the general approach? The intent is to be able to allow the HA integration to show current configuration values and to set them

MartinHjelmare:
I don't have the full picture clear.

One thing I reacted on was the Enum that had a type called enum. A bit weird and funny, I thought. 😄

MartinHjelmare:
Are those all the configuration types that are needed?

raman325:
yeah I paused when I wrote that enum... We could switch it to "choice" so that it's less confusing.

As far as I am aware in my conversations with AlCalzone, yes. Here are the types that are supported: https://github.com/zwave-js/node-zwave-js/blob/master/packages/zwave-js/src/lib/commandclass/ConfigurationCC.ts#L70-L75

Unsigned and SignedInt are ranges, Enumerator is enum, and BitField is enum when the bits are broken out in the device config (example: ). When the bits aren't broken out, the value is returned as an int (base 10 version of the bit settings). If a range is provided, it will act like a range, otherwise it will be `UNDEFINED`.

raman325:
any thoughts on how to address the mypy errors? I'm not sure how to address them without adding ignore statements

zwave_js_server/model/node.py
```diff
@@ -66,9 +69,13 @@ def __init__(self, client: "Client", data: NodeDataType) -> None:
         super().__init__()
         self.client = client
         self.data = data
-        self.values = {
-            get_value_id(self, val): Value(self, val) for val in data["values"]
-        }
+        self.values = {}
```
MartinHjelmare, line 72: Try this:
```suggestion
        self.values: Dict[str, Union[ConfigurationValue, Value]] = {}
```

zwave_js_server/model/node.py
```diff
@@ -234,6 +241,25 @@ def interview_attempts(self) -> Optional[int]:
...
+            for value_id, value in self.values.items()
+            if value.command_class == command_class
+            and (endpoint is None or value.endpoint == endpoint)
+        }
+
+    def get_configuration_values(
+        self, endpoint: int = None
+    ) -> Dict[str, ConfigurationValue]:
+        """Return all configuration values for a node."""
+        return self.get_command_class_values(
```
MartinHjelmare, line 259: I think we need to cast the return value type here.

zwave_js_server/model/node.py
```diff
@@ -246,6 +272,19 @@ async def async_set_value(self, val: Union[Value, str], new_value: Any) -> bool:
         # a value may be specified as value_id or the value itself
         if not isinstance(val, Value):
```
MartinHjelmare, line 273: This should handle subclasses of `Value` too, so it should still work.

zwave_js_server/model/node.py
```diff
@@ -246,6 +272,19 @@ async def async_set_value(self, val: Union[Value, str], new_value: Any) -> bool:
...
             val = self.values[val]
+
+        if not val.metadata.writeable:
+            raise UnwriteableValue
+
+        if (
+            isinstance(val, ConfigurationValue)
```
MartinHjelmare, line 280: We may need to add `ConfigurationValue` to the union in the method signature.

MartinHjelmare: Although typing should handle subclasses, so I'm not sure.

zwave_js_server/model/node.py
raman325: I have done that but I am still getting type errors as follows:
```
zwave_js_server/model/node.py:78: error: Incompatible types in assignment (expression has type "Value", target has type "ConfigurationValue")
zwave_js_server/model/node.py:386: error: Incompatible types in assignment (expression has type "Value", target has type "ConfigurationValue")
zwave_js_server/model/node.py:393: error: Incompatible types in assignment (expression has type "Value", target has type "ConfigurationValue")
```

One solution that works is if we create an `__init__` for `ConfigurationValue` that accepts a `Value` and creates a new object entirely. We can initialize each in the `get_configuration_values` method instead of trying to set the right class type during initialization of the node. Thoughts on this approach? It does eliminate the mypy errors but it feels wasteful to create a copy of the object.

zwave_js_server/model/node.py
raman325: Mapping resulted in the following:
```
zwave_js_server/model/node.py:386: error: Unsupported target for indexed assignment ("Mapping[str, Union[Value, ConfigurationValue]]")
zwave_js_server/model/node.py:393: error: Unsupported target for indexed assignment ("Mapping[str, Union[Value, ConfigurationValue]]")
zwave_js_server/model/node.py:408: error: "Mapping[str, Union[Value, ConfigurationValue]]" has no attribute "pop"
```

but switching to Dict eliminated the errors. Forgot that I needed to type `self.values`

raman325:
should `async_set_value` guard against invalid configuration values? E.g. providing a value for a `RANGE` configuration value that is outside the range, or providing an invalid state for an `ENUMERATED` value

raman325:
I went ahead and added this, but we can drop this last commit if needed

zwave_js_server/model/node.py
```diff
@@ -279,14 +280,36 @@ async def async_set_value(self, val: Union[Value, str], new_value: Any) -> bool:
...
+                    msg += f"Min: {min}, "
+                if max is not None:
+                    msg += f"Max: {max}"
+                raise InvalidNewValue(
+                    f"Must provide a value within the target range ({msg.strip()})"
+                )
+
+            if (
+                val.type == ConfigurationValueType.ENUMERATED
+                and str(new_value) != val.metadata.states
```
MartinHjelmare, line 308: Isn't `val.metadata.states` a dict?

raman325: thanks, it's late

MartinHjelmare (approved):
Nice!

MartinHjelmare:
Should we merge?

raman325:
I'm done with my changes so I think so

PR #108, opened 2021-02-08
Parse non-ASCII lock usercodes

raman325:
Per a conversation with AlCalzone in Discord, usercodes will sometimes be returned as non-ASCII buffers. Per this conversation: https://discord.com/channels/330944238910963714/800356888827002880/807919548081766400 we should handle parsing in the case where such a buffer gets returned.

All three statements in the try block can raise an Exception which is why they are all in the try block.

### DIFF
--- a/zwave_js_server/util/lock.py
+++ b/zwave_js_server/util/lock.py
@@ -65,7 +65,9 @@ def _get_code_slots(
         if include_usercode:
             if not value.value:
                 slot[ATTR_USERCODE] = None
-            elif value.value[0] == "{":
+            elif value.value[0] != "{":
+                slot[ATTR_USERCODE] = value.value
+            else:
                 try:
                     parsed_val = json.loads(value.value)
                     if (
@@ -79,8 +81,6 @@ def _get_code_slots(
                     raise UnparseableValue(
                         f"Unparseable value for code slot {code_slot}: {value.value}"
                     )
-            else:
-                slot[ATTR_USERCODE] = value.value
 
         slots.append(slot)
         code_slot += 1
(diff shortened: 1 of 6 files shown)


raman325:
we could use voluptuous schema validation instead of doing this check: https://github.com/home-assistant-libs/zwave-js-server-python/pull/108/files#diff-1d10555d69219e457f1aaa9d570be975a4d2fa9ffbba9cbe394937643b3e4eadR73-R78

But I wasn't sure if it made sense to add another dependency just for this

raman325:
also per the Discord conversation, this has been scene in the User Code CC and Notification CC. Should we move this validation/parsing logic to the node instantiation?

zwave_js_server/util/lock.py
```diff
@@ -52,15 +53,34 @@ def _get_code_slots(
...
+                try:
+                    parsed_val = json.loads(value.value)
+                    if (
+                        "type" not in parsed_val
+                        or parsed_val["type"] != "Buffer"
+                        or "data" not in parsed_val
+                    ):
+                        raise ValueError("JSON string does not match expected schema")
+                    slot[ATTR_USERCODE] = "".join([chr(x) for x in parsed_val["data"]])
+                except (ValueError, json.decoder.JSONDecodeError) as err:
```
MartinHjelmare, line 80: `JSONDecodeError` is a subclass of `ValueError`.
```suggestion
                except ValueError as err:
```

MartinHjelmare:
I think the parsing should be done when getting the data before building the model representation. The util tools should just enable easier collecting of values.

MartinHjelmare (approved):
Looks good!

We may want to differentiate the exception handling in the integration now so we don't reload the integration on value parsing errors.

raman325:
Good idea. I can make the change on the integration side when we release this

PR #109, opened 2021-02-09
Handle TypeError on connect

MartinHjelmare:
- If we connect before the server is ready, we may get a TypeError when trying to parse the websocket message data as json.
- Refactor the websocket message reception to always use `client.receive` instead of `client.receive_json`. This way we make sure to handle messages with incorrect type.

fixes https://github.com/home-assistant/core/issues/46180

### DIFF
--- a/zwave_js_server/client.py
+++ b/zwave_js_server/client.py
@@ -14,6 +14,7 @@ from .const import MAX_SERVER_VERSION, MIN_SERVER_VERSION
 from .event import Event
 from .exceptions import (
     CannotConnect,
+    ConnectionClosed,
     ConnectionFailed,
     FailedCommand,
     InvalidMessage,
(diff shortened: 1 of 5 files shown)


zwave_js_server/client.py
```diff
@@ -83,6 +83,7 @@ async def connect(self) -> None:
         except (
             client_exceptions.WSServerHandshakeError,
             client_exceptions.ClientError,
+            TypeError,  # TypeError if the client receives non string
```
balloob, line 86: I think the more correct way is to move fetching version info out of this try…catch block and make a helper `receive_json_or_raise` with [the same logic we use in `listen`](). We can then use that for version message, state message and for receiving events.

https://github.com/home-assistant-libs/zwave-js-server-python/blob/master/zwave_js_server/client.py#L152-L172

zwave_js_server/client.py
```diff
@@ -203,6 +186,33 @@ async def disconnect(self) -> None:
         await self._client.close()
         await self._shutdown_complete_event.wait()
 
+    async def _receive_json_or_raise(self) -> dict:
+        """Receive json or raise InvalidMessage."""
```
MartinHjelmare, line 190:
```suggestion
        """Receive json or raise."""
```

balloob (approved):

PR #118, opened 2021-02-12
Rename type attribute for configuration values

raman325:
In reviewing https://github.com/home-assistant/core/pull/46463 I realized that the new `type` attribute for `ConfigurationValue`s may confuse people since `metadata` also has a `type` property. Renaming the attribute so that it is more clear


MartinHjelmare (approved):
Thanks!

PR #119, opened 2021-02-12
Add description property to ValueMetadata

cgarwood:
ValueMetadata class was missing the `description` property


MartinHjelmare:
Can we add a test using a dump that checks that the value metadata returns the correct values from properties?

MartinHjelmare (approved):
Looks good!

PR #120, opened 2021-02-15
Add driver support for update_log_config command

raman325:
Adds library support for the command added here: https://github.com/zwave-js/zwave-js-server/pull/116

The underlying driver function (`updateLogConfig`) was added in `node-zwave-js` version 6.3.0 so this would require the minimum server version to be bumped to beta7 I think. Alternatively we can add some logic in Home Assistant to only register the service if the server version is at least beta7


zwave_js_server/model/driver.py
```diff
@@ -29,3 +30,14 @@ def receive_event(self, event: Event) -> None:
 
     def handle_all_nodes_ready(self, event: Event) -> None:
         """Process a driver all nodes ready event."""
+
+    async def async_update_log_config(self, log_config: LogConfig) -> bool:
```
MartinHjelmare, line 34: I think the parameter should be a dataclass.

test/model/test_driver.py
```diff
@@ -20,3 +21,21 @@ def test_from_state():
...
+
+
+async def test_update_log_config(driver, uuid4, mock_command):
+    """Test set value."""
+    ack_commands = mock_command(
+        {"command": "update_log_config", "config": {"level": 0}},
+        {"success": True},
+    )
+    assert await driver.async_update_log_config(
+        log_config_pkg.LogConfig(level=LogLevel.ERROR)
```
MartinHjelmare, line 33: Let's set all the config attributes so we test everything.

zwave_js_server/const.py
```diff
@@ -7,6 +7,21 @@
...
 
 
+class LogLevel(IntEnum):
+    """Enum for log levels used by node-zwave-js."""
+
+    # https://github.com/zwave-js/node-zwave-js/blob/master/packages/core/src/log/shared.ts#L12
+    # https://github.com/winstonjs/triple-beam/blame/master/config/npm.js#L14
+    ERROR = 0
+    WARN = 1
+    WARNING = WARN
```
balloob, line 17: According to the above linked NPM levels, `WARNING` is not a thing.

zwave_js_server/model/driver.py
```diff
@@ -29,3 +32,18 @@ def receive_event(self, event: Event) -> None:
...
+        """Update log config for driver."""
+        result = await self.client.async_send_command(
+            {
+                "command": "update_log_config",
+                "config": {
+                    snake_to_camel_case(field.name): getattr(log_config, field.name)
```
balloob, line 42: This seems a little hacky. Shouldn't we just require the `LogConfig` class to use the correct formatting that ZJS Server expects? And otherwise maybe just manually rewrite `logToFile` and `forceConsole` instead of all this logic for 2 fields 😬

zwave_js_server/const.py
raman325: I added that in because they are used interchangeably in python, but python doesn't have a `silly` mode so I suppose I should defer to the linked module

zwave_js_server/model/driver.py
raman325: yep we can do that, I over thought this one and was thinking that this could be used elsewhere, but better to make things more explicit so that we can catch if/when the model changes

zwave_js_server/model/log_config.py
```diff
@@ -0,0 +1,50 @@
...
+            "enabled": self.enabled,
+            "level": self.level,
+            "logToFile": self.log_to_file,
+            "filename": self.filename,
+            "forceConsole": self.force_console,
+        }
+        return cast(LogConfigDataType, {k: v for k, v in data.items() if v is not None})
+
+    @staticmethod
+    def from_dict(data: LogConfigDataType) -> "LogConfig":
```
MartinHjelmare, line 42: We don't use this. Do we have plans already for it?

raman325: It's used in the next PR (add support to `get_log_config`)

MartinHjelmare (approved):
Looks good!

PR #121, opened 2021-02-15
Add driver support for get_log_config command

raman325:
Follow on to https://github.com/home-assistant-libs/zwave-js-server-python/pull/120 and https://github.com/zwave-js/node-zwave-js/pull/1770 to add library support for a new command

### DIFF
--- a/zwave_js_server/model/log_config.py
+++ b/zwave_js_server/model/log_config.py
@@ -1,5 +1,4 @@
 """Provide a model for the log config."""
-from __future__ import annotations
 from dataclasses import dataclass
 from typing import Optional, TypedDict, cast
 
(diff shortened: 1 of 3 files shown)


zwave_js_server/model/log_config.py
```diff
@@ -1,4 +1,5 @@
 """Provide a model for the log config."""
+from __future__ import annotations
```
MartinHjelmare, line 2: We can update `from_dict` to return type `LogConfig` with this.

raman325: Oh I meant to not include that and stick with the quoted type since that's used elsewhere. Which would we prefer? I'd like to make it consistent across the board and I don't have a particular preference for one way or another

MartinHjelmare: We can keep it quoted for now and remove this. We can update everything later. 👍

MartinHjelmare (approved):
Looks good!

Issue #122, opened 2021-02-17
ISSUE - Using property key name in value ID means we are missing values in the node's value dictionary


raman325:
While working on config parameter support, we noticed that all values that have a property key don't always have a property key name. This means that partial parameters are probably not working as expected:
```json
{
    "commandClassName": "Configuration",
    "commandClass": 112,
    "endpoint": 0,
    "property": 101,
    "propertyKey": 1,
    "propertyName": "Group 1: Send battery reports",
    "metadata": {
        "type": "number",
        "readable": true,
        "writeable": true,
        "valueSize": 4,
        "min": 0,
        "max": 1,
        "default": 1,
        "format": 0,
        "allowManualEntry": true,
        "label": "Group 1: Send battery reports",
        "description": "Include battery information in periodic reports to Group 1",
        "isFromConfig": true
    },
    "value": 1
},
{
    "commandClassName": "Configuration",
    "commandClass": 112,
    "endpoint": 0,
    "property": 101,
    "propertyKey": 16,
    "propertyName": "Group 1: Send ultraviolet reports",
    "metadata": {
        "type": "number",
        "readable": true,
        "writeable": true,
        "valueSize": 4,
        "min": 0,
        "max": 1,
        "default": 1,
        "format": 0,
        "allowManualEntry": true,
        "label": "Group 1: Send ultraviolet reports",
        "description": "Include ultraviolet information in periodic reports to Group 1",
        "isFromConfig": true
    },
    "value": 1
},
{
    "commandClassName": "Configuration",
    "commandClass": 112,
    "endpoint": 0,
    "property": 101,
    "propertyKey": 32,
    "propertyName": "Group 1: Send temperature reports",
    "metadata": {
        "type": "number",
        "readable": true,
        "writeable": true,
        "valueSize": 4,
        "min": 0,
        "max": 1,
        "default": 1,
        "format": 0,
        "allowManualEntry": true,
        "label": "Group 1: Send temperature reports",
        "description": "Include temperature information in periodic reports to Group 1",
        "isFromConfig": true
    },
    "value": 1
},
{
    "commandClassName": "Configuration",
    "commandClass": 112,
    "endpoint": 0,
    "property": 101,
    "propertyKey": 64,
    "propertyName": "Group 1: Send humidity reports",
    "metadata": {
        "type": "number",
        "readable": true,
        "writeable": true,
        "valueSize": 4,
        "min": 0,
        "max": 1,
        "default": 1,
[...]

PR #123, opened 2021-02-17
Use both property key and property key name in Value ID

raman325:
## Breaking Change
The format of value ID's has changed, as has the parameters for `get_value_id`. 

Resolves https://github.com/home-assistant-libs/zwave-js-server-python/issues/122 (see issue for more context)

### DIFF
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -209,9 +209,9 @@ def driver_fixture(client, controller_state):
     return Driver(client, controller_state)
 
 
-@pytest.fixture(name="node")
-def node_fixture(driver, multisensor_6_state):
-    """Return a node instance with a supporting client."""
+@pytest.fixture(name="multisensor_6")
+def multisensor_6_fixture(driver, multisensor_6_state):
+    """Mock a multisensor 6 node."""
     node = Node(driver.client, multisensor_6_state)
     driver.controller.nodes[node.node_id] = node
     return node
@@ -221,14 +221,6 @@ def node_fixture(driver, multisensor_6_state):
 def lock_schlage_be469_fixture(driver, lock_schlage_be469_state):
     """Mock a schlage lock node."""
     node = Node(driver.client, lock_schlage_be469_state)
-    driver.controller.nodes[node.node_id] = node
-    return node
-
-
-@pytest.fixture(name="multisensor_6")
-def multisensor_6_fixture(driver, multisensor_6_state):
-    """Mock a multisensor 6 node."""
-    node = Node(driver.client, multisensor_6_state)
     driver.controller.nodes[node.node_id] = node
     return node
 
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -225,6 +225,14 @@ def lock_schlage_be469_fixture(driver, lock_schlage_be469_state):
     return node
 
 
+@pytest.fixture(name="multisensor_6")
+def multisensor_6_fixture(driver, multisensor_6_state):
+    """Mock a multisensor 6 node."""
+    node = Node(driver.client, multisensor_6_state)
+    driver.controller.nodes[node.node_id] = node
+    return node
+
+
 @pytest.fixture(name="climate_radio_thermostat_ct100_plus")
 def climate_radio_thermostat_ct100_plus_fixture(
     driver, climate_radio_thermostat_ct100_plus_state
--- a/test/model/test_node.py
+++ b/test/model/test_node.py
@@ -70,8 +70,8 @@ def test_from_state():
 async def test_values_without_property_key_name(multisensor_6):
     """Test that values with property key and without property key name can be found."""
     node = multisensor_6
-    assert "52-112-00-101-1-00" in node.values
-    assert "52-112-00-101-16-00" in node.values
+    assert "52-112-0-101-1-00" in node.values
+    assert "52-112-0-101-16-00" in node.values
 
 
 async def test_command_class_values(climate_radio_thermostat_ct100_plus):
@@ -87,13 +87,13 @@ async def test_command_class_values(climate_radio_thermostat_ct100_plus):
         assert isinstance(value, ConfigurationValue)
 
     with pytest.raises(UnwriteableValue):
-        await node.async_set_value("13-112-00-2-00-00", 1)
+        await node.async_set_value("13-112-0-2-00-00", 1)
 
     with pytest.raises(InvalidNewValue):
-        await node.async_set_value("13-112-00-1-00-00", 5)
+        await node.async_set_value("13-112-0-1-00-00", 5)
 
     with pytest.raises(InvalidNewValue):
-        await node.async_set_value("13-112-00-10-00-00", 200)
+        await node.async_set_value("13-112-0-10-00-00", 200)
 
 
 async def test_set_value(multisensor_6, uuid4, mock_command):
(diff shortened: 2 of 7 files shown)


test/conftest.py
```diff
@@ -225,6 +225,14 @@ def lock_schlage_be469_fixture(driver, lock_schlage_be469_state):
     return node
 
 
+@pytest.fixture(name="multisensor_6")
```
MartinHjelmare, line 228: This fixture is already available as `node`. We can rename that one instead.

zwave_js_server/model/value.py
```diff
@@ -47,8 +47,12 @@ def get_value_id(node: "Node", event_data: ValueDataType) -> str:
     command_class = event_data["commandClass"]
     endpoint = event_data.get("endpoint") or "00"
     property_ = event_data["property"]
-    property_key_name = event_data.get("propertyKeyName") or "00"
-    return f"{node.node_id}-{command_class}-{endpoint}-{property_}-{property_key_name}"
+    # Some values that have a property key don't have a property key name, so we will
+    # fall back to property key if it exists
+    property_key = (
```
MartinHjelmare, line 52: I'm thinking it's best to be consistent and always include both property key and property key name. Is there a downside to that? Then we know what each position represents.

raman325: The main downside is that all places where we call `get_value_id` have to be updated to include both, otherwise no

raman325:
I'm going to update the scope of this PR to add some changes related to setting configuration parameters since it's all related

zwave_js_server/model/value.py
MartinHjelmare: Do you mean if the return value is parsed?

raman325: sorry what I meant was that we would have to update `entity.get_zwave_value` to accept the new parameter and make the corresponding change everywhere. We'd also have to know both the property key and the key name for all values.

raman325: Or we update our approach for value IDs so that we can search for a value based on partial information

MartinHjelmare: I think `get_value_id` should not accept an (event) dictionary but named arguments.

raman325: Whether it comes in via an event dictionary or as named parameters, if we want to include both parameters in the id, anywhere where we need to get a value id we would have to supply both

raman325: What if instead of calling `get_value_id` directly in Home Assistant, we had a method that allowed a subset of parameters to be provided and looped through the Values until it found a matching one, returning the corresponding `value_id`? That would solve the problem of putting all of the parameters in the value_id and not requiring HA to know every parameter in order to get the Value

MartinHjelmare: `propertyKey` is optional in Z-Wave JS just as `propertyKeyName` so the former would only be required when it's present and matters. Currently we supply `propertyKeyName` in 8 places in Home Assistant out of 30 calls to `get_zwave_value`.

The question about who should know what properties are needed is separate from changing this function. I'm not sure about abstracting that more. If you know what value you want you should know how to identify it, I think.

raman325: I see, so you are suggesting we:
1) Change get_value_id to accepted named arguments instead of a dict
2) Add propertyKey and propertyKeyName to the value ID string
2) Change those 8 places to pass both propertyKey and propertyKeyName

Is that correct?

MartinHjelmare: Yes, that's basically my suggestion. 👍

zwave_js_server/model/value.py
```diff
@@ -42,17 +42,31 @@ class ValueDataType(TypedDict, total=False):
...
+        val.get("propertyKeyName"),
     )
-    return f"{node.node_id}-{command_class}-{endpoint}-{property_}-{property_key}"
+
+
+def get_value_id(
+    node: "Node",
+    command_class: CommandClass,
+    property_: Union[str, int],
+    endpoint: int = None,
```
MartinHjelmare, line 61: If we pass None as default the parameter is optional.
```suggestion
    endpoint: Optional[int] = None,
```

zwave_js_server/model/value.py
```diff
@@ -42,17 +42,31 @@ class ValueDataType(TypedDict, total=False):
...
+    property_key: Union[str, int] = None,
+    property_key_name: str = None,
+) -> str:
+    """Return ID of value."""
+    endpoint = endpoint or "00"
```
MartinHjelmare: We should probably check for None explicitly to allow 0 as endpoint.

zwave_js_server/model/value.py
```diff
@@ -42,17 +42,31 @@ class ValueDataType(TypedDict, total=False):
...
+    property_key = property_key or "00"
```
MartinHjelmare, line 67: Here too.

zwave_js_server/model/value.py
```diff
@@ -42,17 +42,31 @@ class ValueDataType(TypedDict, total=False):
...
-def get_value_id(node: "Node", event_data: ValueDataType) -> str:
-    """Return ID of value."""
-    command_class = event_data["commandClass"]
-    endpoint = event_data.get("endpoint") or "00"
-    property_ = event_data["property"]
-    # Some values that have a property key don't have a property key name, so we will
-    # fall back to property key if it exists
-    property_key = (
-        event_data.get("propertyKeyName") or event_data.get("propertyKey") or "00"
+def get_value_id_from_dict(node: "Node", val: ValueDataType):
```
MartinHjelmare: Do we want to expose this helper or keep it private?

raman325: We can keep it private since it's only useful for receiving data from the server

zwave_js_server/model/value.py
```diff
@@ -42,13 +42,35 @@ class ValueDataType(TypedDict, total=False):
...
+        val.get("propertyKey"),
+        val.get("propertyKeyName"),
+    )
+
+
+def get_value_id(
+    node: "Node",
+    command_class: int,
+    property_: Union[str, int],
+    endpoint: Optional[Union[str, int]] = None,
```
MartinHjelmare, line 61: According to the zwave-js docs endpoint is an optional number. Is that correct?
```suggestion
    endpoint: Optional[int]] = None,
```

raman325: yes, this was a `mypy` error I was resolving but I can fix it differently

test/model/test_node.py
```diff
@@ -205,7 +217,7 @@ async def test_get_value_metadata(node, uuid4, mock_command):
         },
     )
 
-    value_id = "52-32-00-targetValue-00"
+    value_id = "52-32-0-targetValue-00-00"
```
MartinHjelmare, line 220: Do you think the endpoint fix has been a problem somewhere in production code?

raman325: I don't believe the endpoint change has caused problems because it was consistent (values on endpoint 0 would always use `00`, and in HA, we fallback on primary value endpoint if it's not provided), but it is a breaking change that will need to be accounted for when we do a version bump

MartinHjelmare (approved):
Great!

test/model/test_node.py
MartinHjelmare: We changed the interface of this function at least so it's good to call it out.

PR #124, opened 2021-02-17
Add async_set_config_parameter function

raman325:
Create a new `async_set_config_parameter` node util function as part of trying to make https://github.com/home-assistant/core/pull/46673 happen


zwave_js_server/util/node.py
```diff
@@ -0,0 +1,106 @@
...
+        )
+
+    # If needed, convert a state label to its key. We know the state exists because
+    # of the validation above.
+    if isinstance(new_value, str):
+        new_value = int(
+            next(
+                key
+                for key in zwave_value.metadata.states
+                if zwave_value.metadata.states[key] == new_value
```
MartinHjelmare, line 78:
```suggestion
                for key, label in zwave_value.metadata.states.items()
                if label == new_value
```

zwave_js_server/util/node.py
```diff
@@ -0,0 +1,106 @@
...
+            bounds.append(f"Min: {min_}")
+        if max_ is not None:
+            bounds.append(f"Max: {max_}")
+        raise InvalidNewValue(
+            f"Must provide a value within the target range ({', '.join(bounds)})"
+        )
+
+    # Finally attempt to set the value and return the Value object if successful
+    if not await node.async_set_value(zwave_value, new_value):
+        return None
```
MartinHjelmare, line 104: I'm thinking if we should raise an error in this case instead?

I also realize now that we don't check this return value in the core integration. We probably should do that.

Here's when it can return False:
> It resolves to false if any of the following conditions are met:
>
> - The setValue API is not implemented in the required Command Class
> - The required Command Class is not implemented in this library yet
> - The API for the required Command Class is not implemented in this library yet

raman325: Should we raise errors for each condition in `async_set_value` instead? Can we detect each of these scenarios?

MartinHjelmare: It's just a general False return value we get from the server. I don't think we can detect details more than that.

https://zwave-js.github.io/node-zwave-js/#/api/node?id=setvalue

raman325: OK 👍🏾

raman325: do we want to raise an error if false? or just add handling on the integration side?

MartinHjelmare: The return value of the function will be cleaner if we raise an error.

Long term we can change `node.async_set_value` to raise this error instead.

MartinHjelmare (approved):
Looks good!

zwave_js_server/util/node.py
```diff
@@ -0,0 +1,110 @@
...
+from ..model.node import Node
+from ..model.value import ConfigurationValue, get_value_id
+
+
+async def async_set_config_parameter(
+    node: Node,
+    new_value: Union[int, str],
+    property_or_property_name: Union[int, str],
+    property_key: Optional[Union[int, str]] = None,
+) -> Optional[ConfigurationValue]:
```
MartinHjelmare, line 16:
```suggestion
) -> ConfigurationValue:
```

PR #125, opened 2021-02-19
Add ColorSwitchCC enum to prep for get_value_id changes

raman325:
The light platform is the only other platform aside from climate that uses `property_key_name` to get a value so we need an enum to be able to use both `property_key` and `property_key_name`

To make these property_key enums easier to use, I added a new dataclass that makes the property_key and property_key_name named attributes


MartinHjelmare (approved):
Looks good!

PR #126, opened 2021-02-19
Don't store unparseable values

raman325:
~~Reverting a change I made earlier. Thinking about it again, it doesn't make sense to raise an exception when a value can't be parsed, it just means that the library is not smart enough (yet!) to parse it. We should store the value as is if we can't parse it and let the client make a decision on how to handle the value~~

If we can't parse a value, we shouldn't store it

### DIFF
--- a/zwave_js_server/model/node.py
+++ b/zwave_js_server/model/node.py
@@ -1,5 +1,4 @@
 """Provide a model for the Z-Wave JS node."""
-import logging
 from typing import TYPE_CHECKING, Any, Dict, List, Optional, TypedDict, Union, cast
 from zwave_js_server.const import CommandClass
 
@@ -21,8 +20,6 @@ from .value import (
 
 if TYPE_CHECKING:
     from ..client import Client
-
-_LOGGER = logging.getLogger(__name__)
 
 
 class NodeDataType(TypedDict, total=False):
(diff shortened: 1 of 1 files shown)


MartinHjelmare:
How will the client know that the value isn't parsed?

raman325:
good question. We weren't handling this before (current release) so we aren't introducing something new with this PR, just reverting something that would have gone in the new release. I am open to suggestions on how to handle this in a way that lets the client know that the value has not been parsed without preventing initialization of the node

MartinHjelmare:
Maybe just drop the value completely? If it can't be parsed it's no good? We're handling missing values already, right?

zwave_js_server/model/node.py
```diff
@@ -72,10 +75,18 @@ def __init__(self, client: "Client", data: NodeDataType) -> None:
...
-            else:
-                self.values[value_id] = Value(self, val)
+            try:
+                if val["commandClass"] == CommandClass.CONFIGURATION:
+                    self.values[value_id] = ConfigurationValue(self, val)
+                else:
+                    self.values[value_id] = Value(self, val)
+            except UnparseableValue:
+                # If we can't parse the value, don't store it but log it for later
+                _LOGGER.info(
```
MartinHjelmare, line 85: Generally it's nice to avoid side effects in init methods. Maybe we should introduce a classmethod to create the node that does the parsing and change the signature of the Node class to accept the values dict too?

MartinHjelmare: But to avoid side effects in any of the models we'd have to change all the models up to the driver. 🙈

raman325: I'm not following this suggestion. Perhaps it would be easier to remove the logging for now?

MartinHjelmare: I don't like that big change now. Not sure about this.

raman325: Yeah that's a much bigger change. I think for now we can just skip the value and we can plan to address that in the future

MartinHjelmare: Ok 👍

MartinHjelmare (approved):

PR #129, opened 2021-02-21
Handle unknown values by returning None for value.value

raman325:
Sometimes a value is returned as "unknown". We have some values that also don't report a full value, so it makes sense to me to treat them the same way (return None) and let the consumer handle this scenario the same way in each case


MartinHjelmare (approved):
Looks good!

MartinHjelmare:
Should we merge this for now, even though Z-Wave JS may decide to do a solution upstream later?

raman325:
I think we should. Didn't sound like something was going to change upstream for sure, and it's something we should be able to catch in the release notes if it does change in the future

PR #134, opened 2021-02-23
Add hash magic method to driver, controller, nodes and values

raman325:
As part of this discussion: https://github.com/home-assistant/core/pull/46673#discussion_r580845521 we realized that our representation for various models is not unique enough. This make controller, node, and value representations unique so they can be usefully hashed.

Without this, you cannot use the zwave_js.set_config_parameter on entities from two different config entries with the same node ID

### DIFF
--- a/test/model/test_controller.py
+++ b/test/model/test_controller.py
@@ -167,6 +167,11 @@ async def test_stop_exclusion(controller, uuid4, mock_command):
     }
 
 
+async def test_hash(controller):
+    """Test node hash."""
+    assert hash(controller) == hash(controller.home_id)
+
+
 async def test_remove_failed_node(controller, uuid4, mock_command):
     """Test remove failed node."""
     ack_commands = mock_command(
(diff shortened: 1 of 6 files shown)


marcelveldt:
Shouldn't you be using as hash method for this ?

`
def __hash__(self):
    return hash(homeid, nodeid)
`

MartinHjelmare:
The default hash for custom objects isn't based on `__repr__` but this PR may still be a good improvement.

raman325:
> Shouldn't you be adding a hash method for this ?
> 
> ```
> def __hash__(self):
>     return hash(homeid, nodeid)
> ```

OK so keep the representation as is and just improve the hash?

marcelveldt:
I think that is better, future wise but like @MartinHjelmare said there is direct issue atm.

MartinHjelmare:
If we always reload the complete state when starting the listen, is there a case when the nodes native instance id won't be unique?

marcelveldt:
I think it could only be an issue if we instantiate Z-Wave Nodes from the model not driven by the controller logic like for compare actions or something like that. Better be safe than sorry and add in the small has method, creating the has based on home id and node id and it's all good.

raman325:
so hash has been added, revert __repr__ changes or no?

MartinHjelmare (approved):
Looks good!

PR #138, opened 2021-02-24
Fix update_log_config return value

raman325:
The original logic assumed there was a `success` key in the result but that is not the case. We can always return `True` because any errors would not be related to the command itself and therefore the logic to receive a message from server should handle any issues that would arise

### DIFF
--- a/zwave_js_server/model/driver.py
+++ b/zwave_js_server/model/driver.py
@@ -41,7 +41,7 @@ class Driver(EventBase):
     def handle_all_nodes_ready(self, event: Event) -> None:
         """Process a driver all nodes ready event."""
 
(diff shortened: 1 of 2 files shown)


zwave_js_server/model/driver.py
```diff
@@ -43,14 +43,14 @@ def handle_all_nodes_ready(self, event: Event) -> None:
 
     async def async_update_log_config(self, log_config: LogConfig) -> bool:
```
MartinHjelmare, line 44: Should we return `None` implicitly instead then?

raman325: Yes good idea since we aren't using the return anyway

MartinHjelmare (approved):
Looks good!

PR #139, opened 2021-02-24
Add option to send a new value without waiting

marcelveldt:
I'm still not sure if this is really needed but PR it anyway so we can decide what to do.
We got some reports about the integration locking up for some users that can be pinpointed to battery powered devices, for example thermostat (valves).

You send a new value to the device but it will take (a long) time before it's processed and currently we're awaiting the result forever.

According to a little discord discussion with @AlCalzone awaiting the setValue should return the bool indicating the command is processed and the promise resolved to the actual result. So actually I can't explain the lockup because you'd think we get a bool back right away after sending the command (and it's been put in the wake up queue by zwave-js)...

Linked issues:

https://github.com/home-assistant/core/issues/46357

https://github.com/home-assistant/core/issues/46363

https://github.com/home-assistant/core/issues/46648

### DIFF
--- a/zwave_js_server/model/node.py
+++ b/zwave_js_server/model/node.py
@@ -282,9 +282,7 @@ class Node(EventBase):
 
         self.emit(event.type, event.data)
 
-    async def async_set_value(
-        self, val: Union[Value, str], new_value: Any, no_wait: bool = False
-    ) -> Optional[bool]:
+    async def async_set_value(self, val: Union[Value, str], new_value: Any) -> None:
         """Send setValue command to Node for given value (or value_id)."""
         # a value may be specified as value_id or the value itself
         if not isinstance(val, Value):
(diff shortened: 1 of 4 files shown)


zwave_js_server/model/node.py
```diff
@@ -292,14 +294,16 @@ async def async_set_value(self, val: Union[Value, str], new_value: Any) -> bool:
...
-        )
+        args = {
+            "command": "node.set_value",
+            "nodeId": self.node_id,
+            "valueId": val.data,
+            "value": new_value,
+        }
+        if no_wait:
+            await self.client.async_send_command_no_wait(args)
+            return
```
MartinHjelmare, line 305:
```suggestion
            return None
```

MartinHjelmare (approved):

AlCalzone:
> According to a little discord discussion with @AlCalzone awaiting the setValue should return the bool indicating the command is processed and the promise resolved to the actual result. So actually I can't explain the lockup because you'd think we get a bool back right away after sending the command (and it's been put in the wake up queue by zwave-js)...

That's not correct.

The method returns a `Promise<boolean>`. That Promise either:
* resolves to `false` (almost) immediately when the command cannot be queued
* resolves to `true` some time in the future after the command was sent and has been acknowledged by the device
* rejects some time in the future if something went wrong while sending the command

IMO you should not block waiting for a command result, except if you're planning on using the result. And even then, the method that does it should run asynchronously.

marcelveldt:
Thanks! That explains a lot, now it's all clear to me. Thinking about this we should (perhaps) never await setValue. We will also run into trouble with setting configuration values.

marcelveldt:
@MartinHjelmare I've changed the logic according to @AlCalzone's response.
Command where we do not need result are no longer awaited but just fired.

MartinHjelmare (approved):

MartinHjelmare:
Should we merge?

marcelveldt:
> Should we merge?

Yes, and probably release a new version ?
PR for core is on its way

AlCalzone:
> Thinking about this we should (perhaps) never await setValue

Depends on how you call it. You could asynchronously fire off the method that does, but then await inside and react once the call went through.

PR #140, opened 2021-02-24
Adopt changes for DeviceClass dump

marcelveldt:
Adopt the breaking changes in the server regarding DeviceClass and add the additional CommandClass list to the dump.
Bumped server requirement to beta9

Related PR on the server:
https://github.com/zwave-js/zwave-js-server/pull/136


MartinHjelmare (approved):
Looks good!

PR #144, opened 2021-02-25
Add schema version support

marcelveldt:
This belongs to https://github.com/zwave-js/zwave-js-server/pull/139

Instead of relying on server versions we switch to api schema to keep compatibility between clients and server versions.


zwave_js_server/client.py
```diff
@@ -55,8 +54,16 @@ def connected(self) -> bool:
...
-    async def async_send_command(self, message: Dict[str, Any]) -> dict:
+    async def async_send_command(
+        self,
+        message: Dict[str, Any],
+        require_scheme: Optional[int] = None,
+    ) -> dict:
         """Send a command and get a response."""
+        if require_scheme is not None and require_scheme > self.scheme_version:
+            raise InvalidServerVersion(
+                "This command is not available, please update Z-Wave Server to a newer version."
```
MartinHjelmare, line 65: Include what version is required.

zwave_js_server/model/version.py
```diff
@@ -18,4 +20,6 @@ def from_message(cls, msg: dict) -> "VersionInfo":
             driver_version=msg["driverVersion"],
             server_version=msg["serverVersion"],
             home_id=msg["homeId"],
+            min_scheme_version=msg.get("minSchemeVersion", 0),
```
MartinHjelmare, line 23: Let's add a comment about when the scheme versions were included in this message.

zwave_js_server/client.py
marcelveldt: we require a server with supported scheme version, so not a specific version itself.
Should we log that ?

zwave_js_server/model/version.py
marcelveldt: good idea

zwave_js_server/client.py
MartinHjelmare: The minimum required version I mean.

marcelveldt: I've adjusted the message

zwave_js_server/const.py
```diff
@@ -132,7 +132,7 @@ class CommandClass(IntEnum):
     SCREEN_MD = 146
     SECURITY = 152
     SECURITY_2 = 159
-    SECURITY_SCHEME0_MARK = 61696
+    SECURITY_SCHEMA0_MARK = 61696
```
MartinHjelmare, line 135: This should probably not change.

marcelveldt: haha, find replace magic :-)

zwave_js_server/model/version.py
```diff
@@ -18,4 +20,10 @@ def from_message(cls, msg: dict) -> "VersionInfo":
             driver_version=msg["driverVersion"],
             server_version=msg["serverVersion"],
             home_id=msg["homeId"],
+            # schema versions are sent in the response from schema version 1+
+            # this info not present means the server is at schema version 0
+            # at some point in time (when we stop supporting schema version 0),
+            # we could adjust this code and assume the keys are there.
+            min_schema_version=msg.get("minSchemaVersion", 0),
```
MartinHjelmare, line 27: I think we should add a test where we send data from a version 0 server, ie not including these new items. We want to check that we're allowing that here.

marcelveldt: Yes, I did not add any additional tests for added logic yet. I'll see what I can do later today.

balloob (approved):

marcelveldt:
I could use some help writing the tests.

PR #145, opened 2021-02-25
Fix parsing of json value

marcelveldt:
Looks like this slipped through in quality control...

1) The parsing logic was only in the init and never updated on value changed events.
2) It didn't account for a value not matching string.

I've adjusted the outer logic a bit. I do not have these json values on any of my devices to test but I didn't touch the inner logic (that handles the actual json value) itself so it should be fine.

Fixes https://github.com/home-assistant/core/issues/47036
Fixes https://github.com/home-assistant/core/issues/47001


MartinHjelmare:
Can we add a test with the reported issue case as input data?

MartinHjelmare (approved):

PR #146, opened 2021-02-25
Handle buffers returned as dictionary when metadata.type is string

raman325:
Follow on to https://github.com/home-assistant-libs/zwave-js-server-python/pull/145

### DIFF
--- a/zwave_js_server/util/helpers.py
+++ b/zwave_js_server/util/helpers.py
@@ -14,12 +14,10 @@ def is_json_string(value: Any) -> bool:
 
 def parse_buffer(value: Dict[str, Any]) -> str:
     """Parse value dictionary from a buffer data type."""
-    if (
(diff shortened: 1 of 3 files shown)


zwave_js_server/util/helpers.py
```diff
@@ -12,16 +12,18 @@ def is_json_string(value: Any) -> bool:
     return isinstance(value, str) and value.startswith("{") and value.endswith("}")
 
 
+def parse_buffer(value: Dict[str, Any]) -> str:
+    """Parse value dictionary from a buffer data type."""
+    if "type" not in value or value["type"] != "Buffer" or "data" not in value:
```
balloob, line 17:
```suggestion
    if value.get("type") != "Buffer" or "data" not in value:
```

balloob (approved):

PR #147, opened 2021-02-27
Change metadata logic to fix Central Scene events

marcelveldt:
Value notification events may arrive with or without metadata present in the event data.
For example for Central Scene the metadata is sent upfront and stored in our values collection.

This changes the logic a bit so that metadata is properly updated and the combined result sent to the client on the event.

Fixes https://github.com/home-assistant/core/issues/47083

### DIFF
--- a/zwave_js_server/model/value.py
+++ b/zwave_js_server/model/value.py
@@ -129,7 +129,7 @@ class ValueMetadata:
         """Return ccSpecific."""
         return self.data.get("ccSpecific", {})
 
-    def update(self, data):
+    def update(self, data: MetaDataType) -> None:
         """Update data."""
         self.data.update(data)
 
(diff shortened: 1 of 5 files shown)


zwave_js_server/model/value.py
```diff
@@ -129,6 +129,10 @@ def cc_specific(self) -> Dict[str, Any]:
         """Return ccSpecific."""
         return self.data.get("ccSpecific", {})
 
+    def update(self, data):
```
MartinHjelmare, line 132:
```suggestion
    def update(self, data: MetaDataType) -> None:
```

MartinHjelmare:
Please add a test for the node value notification event:
https://app.codecov.io/gh/home-assistant-libs/zwave-js-server-python/compare/147/diff

grizzlyjere:
Hi! Was wondering if this could be included in the HA 2021.3 release?

raman325:
@MartinHjelmare are you comfortable with the tests that I added here? I had to fake the event data to test the scenarios but it seemed like a reasonable way to validate the logic.

marcelveldt:
Thanks for finishing the PR @raman325 !

raman325 (approved):

PR #150, opened 2021-03-02
Support string for log level in get_log_config

raman325:
From the conversation here: https://discord.com/channels/330944238910963714/800356888827002880/816067582812618782

We should expect `get_log_config` to return a string in the future. This change supports both the current version and the future version.

### DIFF
--- a/test/model/test_driver.py
+++ b/test/model/test_driver.py
@@ -87,7 +87,7 @@ async def test_update_log_config(driver, uuid4, mock_command):
 
 
 async def test_get_log_config(driver, uuid4, mock_command):
-    """Test set value."""
+    """Test get_log_config."""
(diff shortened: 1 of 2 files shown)


raman325:
Thinking about this a bit, if we want to do this right, we shouldn't handle it like this. Instead we should use API versioning and just use the string for gets and updates since that is what's preferred. But that requires more work to be done and coordinated at the same time (changes to server, the library, and HA) versus this change which can be introduced whenever and doesn't affect server or HA. How do we feel about the two approaches?

MartinHjelmare:
I think we should follow the spec defined by Z-Wave JS. So I don't think we should double up but use the schema versions to adapt to spec changes.

PR #151, opened 2021-03-03
Assume value is on endpoint 0 when value notification is received without an endpoint and add find_value helper

raman325:
https://github.com/home-assistant-libs/zwave-js-server-python/pull/147 was an attempt at fixing the issue users are facing in beta where the value notification is missing parameters because it is missing metadata (https://github.com/home-assistant/core/issues/47083 for context). The previous fix works when the endpoint is provided in the value notification, but from the examples posted (https://discord.com/channels/330944238910963714/800356888827002880/816498596924489768, https://github.com/home-assistant/core/issues/47083#issuecomment-789355704, and https://github.com/home-assistant/core/issues/47083#issuecomment-789357173) it appears that value notification events don't include an endpoint.

@kpine tested an earlier version of zwave-js-server and this appears to have always been the case. What changed is that we are now explicitly checking for `endpoint is None` before using `00` for the endpoint in the value ID, whereas before, `None` or `0` for the endpoint will result in `00` in the value ID. Reverting that change would be another breaking change. My proposed solution is to assume the endpoint is 0 if the endpoint is not provided, which from the docs () seems reasonable. I had originally implemented a value search when the endpoint wasn't provided but this solution is more targeted. I left the new helper in place because marcel indicated on Discord that he would have use for this function elsewhere

To test this, all I had to do was remove the endpoint from the `args` of a test event that was already being tested.

### DIFF
--- a/zwave_js_server/model/node.py
+++ b/zwave_js_server/model/node.py
@@ -2,8 +2,7 @@
 from enum import IntEnum
 from typing import TYPE_CHECKING, Any, Dict, List, Optional, TypedDict, Union, cast
 
-from zwave_js_server.const import CommandClass
-
+from ..const import CommandClass
 from ..event import Event, EventBase
 from ..exceptions import UnparseableValue, UnwriteableValue
 from .command_class import CommandClassInfo, CommandClassInfoDataType
@@ -390,6 +389,38 @@ class Node(EventBase):
             require_schema=1,
         )
 
+    def find_value(
+        self,
+        command_class: Optional[int] = None,
+        command_class_name: Optional[str] = None,
+        endpoint: Optional[int] = None,
+        property_: Optional[Union[int, str]] = None,
+        property_name: Optional[str] = None,
+        property_key: Optional[Union[int, str]] = None,
+        property_key_name: Optional[str] = None,
+    ) -> Optional[Union[ConfigurationValue, Value]]:
+        """Find value from input parameters and return if found."""
+        return next(
+            (
+                value
+                for value in self.values.values()
+                if (command_class is None or value.command_class == command_class)
+                and (
+                    command_class_name is None
+                    or value.command_class_name == command_class_name
+                )
+                and (endpoint is None or value.endpoint == endpoint)
+                and (property_ is None or value.property_ == property_)
+                and (property_name is None or value.property_name == property_name)
+                and (property_key is None or value.property_key == property_key)
+                and (
+                    property_key_name is None
+                    or value.property_key_name == property_key_name
+                )
+            ),
+            None,
+        )
+
     def handle_wake_up(self, event: Event) -> None:
         """Process a node wake up event."""
         # pylint: disable=unused-argument
@@ -454,12 +485,26 @@ class Node(EventBase):
     def handle_value_notification(self, event: Event) -> None:
         """Process a node value notification event."""
         # append metadata if value metadata is available
-        value = self.values.get(_get_value_id_from_dict(self, event.data["args"]))
-        if value:
-            value.update(event.data["args"])
-            value_notification = cast(ValueNotification, value)
+        value = None
+        value_notification = ValueNotification(self, event.data["args"])
+
+        # If endpoint isn't provided, we need to search for the Value
+        if value_notification.endpoint is None:
+            value = self.find_value(
+                property_=value_notification.property_,
+                property_name=value_notification.property_name,
+                property_key=value_notification.property_key,
+                property_key_name=value_notification.property_key_name,
+            )
         else:
-            value_notification = ValueNotification(self, event.data["args"])
+            value = self.values.get(
+                _get_value_id_from_dict(self, value_notification.data)
+            )
+
+        if value:
+            value.update(value_notification.data)
+            value_notification = cast(ValueNotification, value)
+
         event.data["value_notification"] = value_notification
 
     def handle_metadata_updated(self, event: Event) -> None:
--- a/zwave_js_server/model/node.py
+++ b/zwave_js_server/model/node.py
@@ -485,26 +485,20 @@ class Node(EventBase):
     def handle_value_notification(self, event: Event) -> None:
         """Process a node value notification event."""
         # append metadata if value metadata is available
-        value = None
-        value_notification = ValueNotification(self, event.data["args"])
+        # If endpoint isn't provided, we can assume the endpoint is 0
+        # https://zwave-js.github.io/node-zwave-js/#/api/valueid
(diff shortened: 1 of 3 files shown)


raman325:
an alternative approach is to assume the endpoint is 0 if it's not provided, but I'm not sure whether or not that's safe

zwave_js_server/model/node.py
```diff
@@ -454,12 +485,26 @@ def handle_value_removed(self, event: Event) -> None:
     def handle_value_notification(self, event: Event) -> None:
         """Process a node value notification event."""
         # append metadata if value metadata is available
-        value = self.values.get(_get_value_id_from_dict(self, event.data["args"]))
+        value = None
+        value_notification = ValueNotification(self, event.data["args"])
+
+        # If endpoint isn't provided, we need to search for the Value
+        if value_notification.endpoint is None:
```
MartinHjelmare, line 492: If endpoint is `None` we know that the endpoint part in the value_id will be `"00"`. Why can't we look up the value with that info?

MartinHjelmare: Is it that some events have the endpoint and some don't over time?

kpine: Maybe clarify with Al? The [docs](https://zwave-js.github.io/node-zwave-js/#/api/valueid?id=valueid) say that a missing endpoint means it's the root endpoint.

> endpoint - (optional) The index of the node's endpoint (sub-device). 0, undefined and **omitting the index** addresses the root device (in fact, these options can be used interchangeably). An index >= 1 addresses one of the single endpoints.

That's at least for setting values, hopefully it's consistent with receiving values?

raman325: the event links to that same part of the docs, so maybe we can simplify this and say that if endpoint isn't included, we use endpoint 0. @MartinHjelmare maybe I am misunderstanding what your original question was but every value should have an endpoint so 00 should never resolve. Perhaps we should default to `0` instead of `00` in the value ID generation. That could be a breaking change tohugh.

jcam: What i saw is that the metadata updates from the interview have endpoint id 0 for the central scene value metadata, while the value notification has no endpoint, so since the default was 00, 00 != 0.

jcam: unsure if the same is true for GE switches with the basic notifications... but maybe those do include an endpoint in the notification

zwave_js_server/model/node.py
```diff
@@ -454,6 +485,11 @@ def handle_value_removed(self, event: Event) -> None:
     def handle_value_notification(self, event: Event) -> None:
         """Process a node value notification event."""
         # append metadata if value metadata is available
+        # If endpoint isn't provided, we can assume the endpoint is 0
+        # https://zwave-js.github.io/node-zwave-js/#/api/valueid
+        if event.data["args"].get("endpoint") is None:
+            event.data["args"]["endpoint"] = 0
+
         value = self.values.get(_get_value_id_from_dict(self, event.data["args"]))
```
MartinHjelmare, line 493: Where are we storing ValueNotification instances in `Node.values`?

MartinHjelmare (commented):
I don't really understand how we want the ValueNotification to work.

zwave_js_server/model/node.py
raman325: We are not storing ValueNotification instances. They get attached to the value notification event by the library before getting passed on to the consumer, e.g. the `zwave_js` integration.

MartinHjelmare: I think what's confusing me is that we're casting the Value instance to a ValueNotification instance. That's not correct. We should always create a ValueNotification instance with the same data as the Value that we're basing the ValueNotification on.

raman325: @marcelveldt do you have any thoughts on this? I can make this change to create a new instance, just not sure if it was an intentional choice to update the original Value rather than creating a new ValueNotification instance?

marcelveldt: You need to fetch the metadata belonging to the Value notification, this will be sent upfront as "value added", "value updated" or "metadata updated" message. We could also revert to the previous behavior where the server collected the metadata on an incoming value notification message...

raman325: That part I understand. It's a matter of order of operations and what object is being updated though. Right now, we are updating the `Value` instance that we retrieve with the value notification data. This means that we are updating the state of the `Value` we are storing. Is this intentional? The alternative approach is to create a new `ValueNotification` instance from the `Value` and update it with the value notification data. This achieves the same thing but doesn't impact the state of the original `Value`

MartinHjelmare: Regardless of the question about updating the existing `Value` instance we should always create a `ValueNotification` instance and set it in the event data. We can't mix types for the same event data.

raman325: `Value` and `ValueNotification` are the same thing internally. Is there a functional issue or is this more of a "right way to do this"?

marcelveldt: It was just a bit future proof at that time but it doesn't make any actual sense to keep it if we're not going to add additional logic.

marcelveldt: > That part I understand. It's a matter of order of operations and what object is being updated though. Right now, we are updating the `Value` instance that we retrieve with the value notification data. This means that we are updating the state of the `Value` we are storing. Is this intentional? The alternative approach is to create a new `ValueNotification` instance from the `Value` and update it with the value notification data. This achieves the same thing but doesn't impact the state of the original `Value`

Uhm, we could probably request the metadata at this point (there's a function for that) and update the value in our states dict. So that way we can guarantee we always have the metadata. That would solve this once and for all.

value notification --> is this value known: use known metadata. value not known and no metadata in notification message, request the metadata from the server, create/update value object and send the complete result to HA

raman325:
@jcam made a good point that #153 would fix the issue because we would handle an endpoint of null as being on endpoint 0. So I am going to close this, if we want to make changes as to how we want to handle value notifications, we can do it in a new PR

zwave_js_server/model/node.py
MartinHjelmare: > `Value` and `ValueNotification` are the same thing internally. Is there a functional issue or is this more of a "right way to do this"?

`ValueNotification` inherits `Value`, but it doesn't mean we can replace a `ValueNotification` instance with a `Value` instance, even if they have the same features currently.

raman325: I see what you mean. I can submit a separate PR to address that

PR #152, opened 2021-03-04
Fix exception

gsemet:
Exception was:
```
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/components/zwave_js/__init__.py", line 374, in client_listen
    await client.listen(driver_ready)
  File "/usr/local/lib/python3.8/site-packages/zwave_js_server/client.py", line 168, in listen
    if not state_msg["success"]:
KeyError: 'success'
```

### DIFF
--- a/zwave_js_server/client.py
+++ b/zwave_js_server/client.py
@@ -165,7 +165,7 @@ class Client:
 
             state_msg = await self._receive_json_or_raise()
 
-            if not state_msg["success"]:
+            if not state_msg.get("success"):
                 await self._client.close()
                 raise FailedCommand(state_msg["messageId"], state_msg["errorCode"])
 


zwave_js_server/client.py
```diff
@@ -165,7 +165,7 @@ async def listen(self, driver_ready: asyncio.Event) -> None:
 
             state_msg = await self._receive_json_or_raise()
 
-            if not state_msg["success"]:
+            if not state_msg.get("success"):
```
MartinHjelmare, line 168: When is the success key missing?

gsemet: it is related to https://github.com/home-assistant/core/issues/47347

MartinHjelmare: If the success key is missing it means we're not receiving the expected message type from the server. We need to look into why this is.

marcelveldt: This is a side effect of sending setValue commands to battery powered nodes (which we no longer await in the python lib). I think we should fix this in the server and not send back an (unexpected) result for commands that we didn't want to await.

kpine: I think the wrong issue was linked, it's https://github.com/home-assistant/core/issues/47357.

I added some logs to that, in my case it had nothing to do with setValue or battery nodes, just a value update message arriving at an unexpected time.

marcelveldt: ah ok, in that case we could it as-is for now.

raman325:
I'm going to close this. As discussed in Discord, we want this to be handled on the server side rather than in the library, and we think we've addressed it with https://github.com/zwave-js/zwave-js-server/pull/152

PR #153, opened 2021-03-04
Update value ID schema from discussion

raman325:
Before I go too much further, I wanted to ensure that everyone agrees with this approach per the discussion in #devs_zwave (https://discord.com/channels/330944238910963714/800356888827002880/816945787766308874). Assuming we are in agreement, I can work on updating tests and the corresponding changes in HA

CC @AlCalzone @marcelveldt @MartinHjelmare

### DIFF
--- a/zwave_js_server/model/value.py
+++ b/zwave_js_server/model/value.py
@@ -192,7 +192,7 @@ class Value:
     @property
     def endpoint(self) -> Optional[int]:
         """Return endpoint."""
-        return self.data.get("endpoint")
+        return self.data.get("endpoint", 0)
 
     @property
     def property_(self) -> Union[str, int]:
--- a/zwave_js_server/model/value.py
+++ b/zwave_js_server/model/value.py
@@ -61,7 +61,7 @@ def get_value_id(
 ) -> str:
     """Return ID of value."""
     # If endpoint is not provided, assume root endpoint
-    endpoint_ = 0 if endpoint is None else endpoint
+    endpoint_ = endpoint or 0
     value_id = f"{node.node_id}-{command_class}-{endpoint_}-{property_}"
     # Property key is only included when it has a value
     if property_key is not None:
--- a/zwave_js_server/model/value.py
+++ b/zwave_js_server/model/value.py
@@ -192,7 +192,7 @@ class Value:
     @property
     def endpoint(self) -> Optional[int]:
         """Return endpoint."""
-        return self.data.get("endpoint", 0)
+        return self.data.get("endpoint")
 
     @property
     def property_(self) -> Union[str, int]:
--- a/zwave_js_server/model/value.py
+++ b/zwave_js_server/model/value.py
@@ -49,7 +49,6 @@ def _get_value_id_from_dict(node: "Node", val: ValueDataType) -> str:
         val["property"],
         val.get("endpoint"),
         val.get("propertyKey"),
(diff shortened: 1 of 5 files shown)


AlCalzone:
Note that using the propertyKey instead of propertyKeyName will result in more "cryptic" IDs - Meters are an example.

Using propertyName instead of property and propertyKeyName instead of propertyKey would yield more human-readable IDs. Both is okay, you just need to be consistent.

raman325:
> Note that using the propertyKey instead of propertyKeyName will result in more "cryptic" IDs - Meters are an example.
> 
> Using propertyName instead of property and propertyKeyName instead of propertyKey would yield more human-readable IDs. Both is okay, you just need to be consistent.

What happens if we don't have a propertyKeyName for a propertyKey? We noticed that being the case for configuration parameters which is why we started down this path in the first place

AlCalzone:
> We noticed that being the case for configuration parameters which is why we started down this path in the first place

You sure? Is that a bug in the server maybe? Most of those are partial parameters and the IDs you see are built from propertyKeyNames:
grafik

AlCalzone:
In any case, you can still use propertyName (+ propertyKeyName) for IDs, but you need to store the original property and propertyKey combination because that is what zwave-js needs to set the values.

raman325:
> In any case, you can still use propertyName (+ propertyKeyName) for IDs, but you need to store a reference to the original property and propertyKey combination because that is what zwave-js needs to set the values.

We store everything in the value object itself, the discussion here started around what we should use for value IDs, and I thought the conclusion was that we wanted to mirror zwavejs2mqtt to make troubleshooting across repos easier? If that's the case, I think we want to stick with property and propertyKey because that is what zjs2m uses.

AlCalzone:
@robertsLando, how does zwavejs2mqtt build its value IDs? property + propertyKey or propertyName + propertyKeyName?

raman325:
I think it would have to use propertyKey, otherwise for partial parameters, multiple partials would have the same value ID

robertsLando:
I use the property key: https://github.com/zwave-js/zwavejs2mqtt/blob/master/lib/ZwaveClient.js#L1213

zwave_js_server/model/value.py
```diff
@@ -195,7 +192,7 @@ def cc_version(self) -> int:
     @property
     def endpoint(self) -> Optional[int]:
         """Return endpoint."""
-        return self.data.get("endpoint")
+        return self.data.get("endpoint", 0)
```
MartinHjelmare, line 195: When is the endpoint missing in the Value data?

raman325: practically speaking, I don't know. The docs implied that there are scenarios where `endpoint` is not provided, and when that is the case, the value is for the root device (0). If we are contending that endpoint is always available, we shouldn't be using a `get` here in the first place

MartinHjelmare: I understood @AlCalzone as it's ok to not provide the endpoint when contructing the ValueId property to use with the zwave-js library. It means the same thing as providing endpoint 0. But we will always get the endpoint back from the library in the state.

It would be good to have this verified.

raman325: I'd like to move forward with this PR since it's causing issues for users. I will revert this change and we can come back to this separately

MartinHjelmare (approved):

PR #154, opened 2021-03-04
Always return a new ValueNotification on value notification events

raman325:
Addresses comment from #151

### DIFF
--- a/zwave_js_server/model/node.py
+++ b/zwave_js_server/model/node.py
@@ -454,12 +454,11 @@ class Node(EventBase):
     def handle_value_notification(self, event: Event) -> None:
         """Process a node value notification event."""
         # append metadata if value metadata is available
+        value_notification = ValueNotification(self, event.data["args"])
         value = self.values.get(_get_value_id_from_dict(self, event.data["args"]))
         if value:
             value.update(event.data["args"])
-            value_notification = cast(ValueNotification, value)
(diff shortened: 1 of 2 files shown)


zwave_js_server/model/node.py
```diff
@@ -454,12 +454,11 @@ def handle_value_removed(self, event: Event) -> None:
     def handle_value_notification(self, event: Event) -> None:
         """Process a node value notification event."""
         # append metadata if value metadata is available
+        value_notification = ValueNotification(self, event.data["args"])
         value = self.values.get(_get_value_id_from_dict(self, event.data["args"]))
         if value:
             value.update(event.data["args"])
```
raman325, line 460: So the only open question that I am not clear on is whether we still need this line

MartinHjelmare (approved):
Looks good!

zwave_js_server/model/node.py
raman325: @marcelveldt sorry to keep tagging you, I just didn't fully understand your comment in the other PR. Should we update the `Value` or should we not? I think from a practical standpoint what we'd be updating is the stored value of the `Value`

raman325: given that this doesn't change functionality by leaving it in, I think we can merge and address this separately if needed

PR #156, opened 2021-03-05
Make sure ValueNotification includes the latest event data

raman325:
In #154 the order of operations for creating the `ValueNotification` was wrong and led to the value that came in the notification being returned as the value that we was previously in the state (see https://github.com/home-assistant/core/issues/47432). This fixes the order of operations.

I could use some help writing some good test cases for this one, I don't think we have enough given all the issues that have occurred as we try to fix this

### DIFF
--- a/zwave_js_server/model/node.py
+++ b/zwave_js_server/model/node.py
@@ -455,13 +455,12 @@ class Node(EventBase):
         """Process a node value notification event."""
         # if value is found, use value data as base and update what is provided
         # in the event, otherwise use the event data
-        if value := self.values.get(
-            _get_value_id_from_dict(self, event.data["args"])
-        ):
+        event_data = event.data["args"]
+        if value := self.values.get(_get_value_id_from_dict(self, event_data)):
             value_notification = ValueNotification(self, value.data)
-            value_notification.update(event.data["args"])
+            value_notification.update(event_data)
         else:
-            value_notification = ValueNotification(self, event.data["args"])
+            value_notification = ValueNotification(self, event_data)
 
         event.data["value_notification"] = value_notification
 
--- a/zwave_js_server/model/node.py
+++ b/zwave_js_server/model/node.py
@@ -460,7 +460,7 @@ class Node(EventBase):
             _get_value_id_from_dict(self, value_notification.data)
         ):
             value_notification.metadata.update(value.metadata.data)
-            if value_notification.endpoint is None:
+            if value_notification.endpoint is None and value.endpoint is not None:
                 value_notification.data["endpoint"] = value.endpoint
         event.data["value_notification"] = value_notification
 
--- a/zwave_js_server/model/node.py
+++ b/zwave_js_server/model/node.py
@@ -455,12 +455,10 @@ class Node(EventBase):
         """Process a node value notification event."""
         # append metadata if value metadata is available
         value_notification = ValueNotification(self, event.data["args"])
-        value = self.values.get(_get_value_id_from_dict(self, value_notification.data))
-        if value:
-            # We update the Value first so that we can merge the Value back into
-            # ValueNotification
-            value.update(value_notification.data)
-            value_notification.update(value.data)
+        if value := self.values.get(
+            _get_value_id_from_dict(self, value_notification.data)
+        ):
+            value_notification.metadata.update(value.metadata.data)
         event.data["value_notification"] = value_notification
 
     def handle_metadata_updated(self, event: Event) -> None:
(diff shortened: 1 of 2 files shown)


raman325:
@kpine can you take a look at this and see if it makes sense since you spotted this immediately?

kpine (approved):
This is working fine for me. I took a slightly different approach, same result.

```python
        if value:
            value.update(event.data["args"])
            value_notification = ValueNotification(self, value.data)
        else:
            value_notification = ValueNotification(self, event.data["args"])
        event.data["value_notification"] = value_notification
```

kpine:
I was wondering, why is `value` modified at all? Isn't it needed just to get the metadata? These notifications are supposed to be events or stateless values, not value updates.

MartinHjelmare:
To test this I'd create a Value instance on the node and then create an event coming in with new data. The `ValueNotification` data should then be created with the Value instance data as base but be updated with the new data from the event. So we need to make sure that we can assert that difference when creating the initial Value and event data

MartinHjelmare:
> I was wondering, why is `value` modified at all? Isn't it needed just to get the metadata? These notifications are supposed to be events or stateless values, not value updates.

Yes, it doesn't sound like the Value state should be updated when reading the docs about ValueNotification event:
https://zwave-js.github.io/node-zwave-js/#/api/node?id=quotvalue-notificationquot

jcam:
Is there any pre-existing method to do this?
            value_notification.update(value.metadata)
I imagine value.metadata doesn't exist yet...

jcam:
ah, perhaps it does...
so maybe this `value_notification.metadata.update(value.metadata)`

MartinHjelmare:
```py
value_notification.metadata.update(value.metadata.data)
```

jcam:
Just tested on my prod setup, this works, and resolves https://github.com/home-assistant/core/issues/47432

> ```python
> value_notification.metadata.update(value.metadata.data)
> ```

raman325:
> To test this I'd create a Value instance on the node and then create an event coming in with new data, that should trigger a ValueNotification event. The `ValueNotification` data should then be created with the Value instance data as base but be updated with the new data from the event. So we need to make sure that we can assert that difference when creating the initial Value and event data

So it looks like my tests are missing a value in the `args` part of the value notification event, and I don't test the value explicitly. If I were to add it with a different value from the `Value`s value, and then check it, would that be sufficient?

raman325:
> > To test this I'd create a Value instance on the node and then create an event coming in with new data, that should trigger a ValueNotification event. The `ValueNotification` data should then be created with the Value instance data as base but be updated with the new data from the event. So we need to make sure that we can assert that difference when creating the initial Value and event data
> 
> So it looks like my tests are missing a value in the `args` part of the value notification event, and I don't test the value explicitly. If I were to add it with a different value from the `Value`s value, and then check it, would that be sufficient?

actually, the `Value` in my test data doesn't even have a value, so if I add a value to the notification, we can guarantee it's coming from the notification event if the `ValueNotification` has a value

raman325:
> ```python
> value_notification.metadata.update(value.metadata.data)
> ```

The problem with this approach is that `ValueNotification.endpoint` is `None`. There are two ways to solve this:
1. Update the entire `value_notification.data`. This **may** have unintended side effects but not sure.
2. Update the endpoint property to return 0 by default. I think this is safe because every value **has** to be on an endpoint, doesn't it?

jcam:
I think the endpoint is not included from the ValueNotification, so the endpoint shouldn't be included in the HA event?

raman325:
> I think the endpoint is not included from the ValueNotification, so the endpoint shouldn't be included in the HA event?

The notification is for a value that **does** have an endpoint. There's a third option, see my upcoming commit

zwave_js_server/model/node.py
```diff
@@ -453,17 +453,21 @@ def handle_value_removed(self, event: Event) -> None:
...
-            value_notification.update(value.data)
-            value.update(event.data["args"])
+        # append metadata if value metadata is available and endpoint
+        # if value notification endpoint isn't provided
+        if value := self.values.get(
+            _get_value_id_from_dict(self, value_notification.data)
+        ):
+            value_notification.metadata.update(value.metadata.data)
+            if value_notification.endpoint is None and value.endpoint is not None:
+                value_notification.data["endpoint"] = value.endpoint
```
MartinHjelmare, line 464: Would it be safe to first update the ValueNotification with the Value data and then update the ValueNotification with the event data?

We'd avoid accessing details of the data on the ValueNotification that way. It's best if we let each model handle how it should update its data.

MartinHjelmare: For reference:
https://github.com/home-assistant-libs/zwave-js-server-python/blob/453318cb752d6be4eb53a243fa595ac5c962da1a/zwave_js_server/model/value.py#L226-L241

MartinHjelmare (approved):

kpine:
Looks good!

FYI, my scene activation and basic set events do include endpoint from the server.

PR #158, opened 2021-03-05
Add selective wait for node command response when device is awake

raman325:
It was easier to write the code to make this proposal than to discuss it on Discord, but this is just an idea that I am looking for feedback on.

Right now, we have updated `node.async_set_value` to support selectively waiting and not waiting for the result of the command, but this distinction has value for other commands as well (like `async_poll_value`). In this PR, I created an internal node function `node._async_send_command` that optionally allows the caller to force a wait or non-wait but can also decide what to do based on the node status.

### DIFF
--- a/zwave_js_server/model/node.py
+++ b/zwave_js_server/model/node.py
@@ -1,6 +1,5 @@
 """Provide a model for the Z-Wave JS node."""
 from enum import IntEnum
-import logging
 from typing import TYPE_CHECKING, Any, Dict, List, Optional, TypedDict, Union, cast
 
 from zwave_js_server.const import CommandClass
@@ -374,13 +373,13 @@ class Node(EventBase):
             "get_defined_value_ids", wait_for_result=True
         )
         if data is not None:
-            return [
-                Value(self, cast(ValueDataType, valueId))
-                for valueId in data["valueIds"]
-            ]
-        else:
             # We should never reach this code
             raise FailedCommand("Command failed", "failed_command")
+
+        return [
+            Value(self, cast(ValueDataType, valueId))
+            for valueId in data["valueIds"]
+        ]
 
     async def async_get_value_metadata(self, val: Union[Value, str]) -> ValueMetadata:
         """Send getValueMetadata command to Node."""
--- a/zwave_js_server/model/node.py
+++ b/zwave_js_server/model/node.py
@@ -362,7 +362,7 @@ class Node(EventBase):
         if result is None:
             return None
 
-        return result["success"]
+        return cast(bool, result["success"])
 
     async def async_refresh_info(self) -> None:
         """Send refreshInfo command to Node."""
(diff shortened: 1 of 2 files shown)


zwave_js_server/model/node.py
```diff
@@ -310,8 +310,36 @@ def receive_event(self, event: Event) -> None:
 
         self.emit(event.type, event.data)
 
+    async def _async_send_command(
+        self,
+        cmd: str,
+        require_schema: Optional[int] = None,
+        wait_for_result: bool = None,
```
MartinHjelmare, line 317: Why put `None` as default? We don't seem to care about it in the method body. It could be `False`.

raman325: That’s a good point, when I had written this originally I expected to use None but it’s not required. Will update it later

raman325: I recall now why we needed None, and I'm glad you mentioned it because I didn't handle it correctly. The idea was that the caller could force a behavior (wait or not wait) with a bool whereas `None` would be based on node status. I'll push a commit that implements this properly

MartinHjelmare:
```suggestion
        wait_for_result: Optional[bool] = None,
```

zwave_js_server/model/node.py
```diff
@@ -310,8 +310,38 @@ def receive_event(self, event: Event) -> None:
...
+        **cmd_kwargs: Any,
+    ) -> Optional[Any]:
```
MartinHjelmare, line 319: Isn't the result always a dict?
```suggestion
    ) -> Optional[dict]:
```

zwave_js_server/model/node.py
```diff
@@ -322,35 +352,31 @@ async def async_set_value(
...
+            "set_value",
+            valueId=val.data,
+            value=new_value,
+            wait_for_result=wait_for_result,
+        )
+
+        if result is None:
+            return None
+
+        return cast(bool, result)
```
MartinHjelmare, line 365: I think this is incorrect. The result is a dict.

zwave_js_server/model/node.py
```diff
@@ -322,35 +352,31 @@ async def async_set_value(
...
+            "set_value",
+            valueId=val.data,
+            value=new_value,
+            wait_for_result=wait_for_result,
+        )
+
+        if result is None:
+            return None
+
+        return isinstance(result, dict) and result.get("success", False)
```
MartinHjelmare, line 365: We know the result type for this command. I think we should blow up if that would be incorrect.

https://github.com/zwave-js/zwave-js-server/blob/c1ddbd1c802712161d8cab2020f1a5a1058d6453/src/lib/node/outgoing_message.ts#L5

```suggestion
        return result["success"]
```

MartinHjelmare (approved):
Looks good!

MartinHjelmare (approved):

PR #163, opened 2021-03-13
Add allowManualEntry metadata

firstof9:
## Proposed change
Allow configuration values with the `allowManualEntry` metadata bool to be set to a range rather than be locked to the values from the `states` list.

Example:
```json
              {
                "endpoint": 0,
                "commandClass": 112,
                "commandClassName": "Configuration",
                "property": 8,
                "propertyKey": 255,
                "propertyName": "LED Strip Effect (Color)",
                "ccVersion": 1,
                "metadata": {
                  "type": "number",
                  "readable": true,
                  "writeable": true,
                  "valueSize": 4,
                  "min": 0,
                  "max": 255,
                  "default": 0,
                  "format": 0,
                  "allowManualEntry": true,
                  "states": {
                    "0": "Red",
                    "21": "Orange",
                    "42": "Yellow",
                    "85": "Green",
                    "127": "Cyan",
                    "170": "Blue",
                    "212": "Violet",
                    "234": "Pink"
                  },
                  "label": "LED Strip Effect (Color)",
                  "isFromConfig": true
                },
                "value": 180
              },
```

### DIFF
--- a/test/util/test_node.py
+++ b/test/util/test_node.py
@@ -33,11 +33,11 @@ async def test_configuration_parameter_values(
     with pytest.raises(NotImplementedError):
         await async_set_config_parameter(node, 1, 2)
 
-    # Test setting an manual entry configuration parameter with an invalid value
+    # Test setting a manual entry configuration parameter with an invalid value
     with pytest.raises(InvalidNewValue):
         await async_set_config_parameter(node_2, "Purple", 8, 255)
 
-    # Test setting an manual entry configuration parameter with an valid value
+    # Test setting a manual entry configuration parameter with a valid value
     ack_commands_2 = mock_command(
         {"command": "node.set_value", "nodeId": node_2.node_id},
         {"success": True},
--- a/test/util/test_node.py
+++ b/test/util/test_node.py
@@ -55,6 +55,18 @@ async def test_configuration_parameter_values(
         "messageId": uuid4,
     }
 
+    await async_set_config_parameter(node_2, "Blue", 8, 255)
+
+    value = node_2.values["31-112-0-8-255"]
+    assert len(ack_commands_2) == 2
+    assert ack_commands_2[1] == {
+        "command": "node.set_value",
+        "nodeId": node_2.node_id,
+        "valueId": value.data,
+        "value": 170,
+        "messageId": uuid4,
+    }
+
     # Test setting an enumerated configuration parameter with an invalid value
     with pytest.raises(InvalidNewValue):
         await async_set_config_parameter(node, 5, 1)
@@ -81,8 +93,8 @@ async def test_configuration_parameter_values(
     )
 
     value = node.values["13-112-0-1"]
-    assert len(ack_commands) == 2
-    assert ack_commands[1] == {
+    assert len(ack_commands) == 3
+    assert ack_commands[2] == {
         "command": "node.set_value",
         "nodeId": node.node_id,
         "valueId": value.data,
--- a/test/util/test_node.py
+++ b/test/util/test_node.py
@@ -37,6 +37,24 @@ async def test_configuration_parameter_values(
     with pytest.raises(InvalidNewValue):
         await async_set_config_parameter(node_2, "Purple", 8, 255)
 
+    # Test setting an manual entry configuration parameter with an valid value
+    ack_commands_2 = mock_command(
+        {"command": "node.set_value", "nodeId": node_2.node_id},
+        {"success": True},
+    )
+
+    await async_set_config_parameter(node_2, 190, 8, 255)
+
+    value = node_2.values["31-112-0-8-255"]
+    assert len(ack_commands_2) == 1
+    assert ack_commands_2[0] == {
+        "command": "node.set_value",
+        "nodeId": node_2.node_id,
+        "valueId": value.data,
+        "value": 190,
+        "messageId": uuid4,
+    }
+
     # Test setting an enumerated configuration parameter with an invalid value
     with pytest.raises(InvalidNewValue):
         await async_set_config_parameter(node, 5, 1)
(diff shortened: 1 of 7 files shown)


firstof9:
@raman325 what do you think?

zwave_js_server/model/value.py
```diff
@@ -126,6 +127,11 @@ def cc_specific(self) -> Dict[str, Any]:
         """Return ccSpecific."""
         return self.data.get("ccSpecific", {})
 
+    @property
+    def allowManualEntry(self) -> Optional[bool]:
```
MartinHjelmare, line 131: Please use snake_case names in the model.

zwave_js_server/model/value.py
```diff
@@ -258,7 +264,7 @@ class ConfigurationValue(Value):
     def configuration_value_type(self) -> ConfigurationValueType:
         """Return configuration value type."""
         if self.metadata.type == "number":
-            if self.metadata.states:
+            if self.metadata.states and not self.metadata.allow_manual_entry:
```
raman325, line 267: This would add support for the scenario you described on Discord, but it might be better to have a new `ConfigurationValueType` for this type of property and then update the `async_set_config_parameter` function to support it (changes required would be minimal).

We should also add tests

firstof9: Tests added.

firstof9: New `ConfigurationValueType` added.

zwave_js_server/util/node.py
```diff
@@ -86,7 +86,11 @@ async def async_set_config_parameter(
     # Validate that new value for range configuration parameter is within bounds
     max_ = zwave_value.metadata.max
     min_ = zwave_value.metadata.min
-    if zwave_value.configuration_value_type == ConfigurationValueType.RANGE and (
+    value_ = (
```
raman325, line 89: We set `min_` and `max_` for convenience because they are referenced multiple times below. `value_` is only referenced once, and it's name implies that it represents a value but it doesn't. Can we collapse this into the if statement and remove `value_` altogether?

raman325 (changes requested):
Currently in this PR we don't guard against someone attempting to set a value using an invalid state label for a manual entry configuration value. We can add that guard here: https://github.com/home-assistant-libs/zwave-js-server-python/blob/cb0b236ed906ff1688bad81f283c58e38ad0de49/zwave_js_server/util/node.py#L73. It's basically the same guard as the one here: https://github.com/home-assistant-libs/zwave-js-server-python/blob/cb0b236ed906ff1688bad81f283c58e38ad0de49/zwave_js_server/util/node.py#L58 except the `ConfigurationValueType` is different and we only want to check the `new_value` against `zwave_value.metadata.states.values()`

zwave_js_server/util/node.py
firstof9: Doing so gave me a linting error, too many bools (6/5).
I'm up for renaming the variable to something else.

firstof9: I've renamed it to `check_` since it's technically checking the value types.

zwave_js_server/util/node.py
```diff
@@ -68,6 +68,18 @@ async def async_set_config_parameter(
...
+    # key or label
+    if (
+        isinstance(new_value, str)
+        and zwave_value.configuration_value_type == ConfigurationValueType.MANUAL_ENTRY
+        and str(new_value) not in [*zwave_value.metadata.states.values()]
+    ):
+        raise InvalidNewValue(
+            "Must provide a value that represents a valid state key or label from "
+            f"{json.dumps(zwave_value.metadata.states)}"
+        )
```
raman325, line 81:
```suggestion
        and str(new_value) not in zwave_value.metadata.states.values()
    ):
        raise InvalidNewValue(
            "Must provide a value that represents a valid state from "
            f"{json.dumps(zwave_value.metadata.states.values())}"
        )
```

zwave_js_server/util/node.py
```diff
@@ -68,6 +68,18 @@ async def async_set_config_parameter(
...
+    # Validate that new value for manual entry configuration parameter is a valid state
+    # key or label
+    if (
+        isinstance(new_value, str)
+        and zwave_value.configuration_value_type == ConfigurationValueType.MANUAL_ENTRY
+        and str(new_value) not in zwave_value.metadata.states.values()
+    ):
+        raise InvalidNewValue(
+            "Must provide a value that represents a valid state from "
+            f"{json.dumps(zwave_value.metadata.states.values())}"
```
raman325, line 80:
```suggestion
            f"{list(zwave_value.metadata.states.values())}"
```

My mistake

test/util/test_node.py
```diff
@@ -21,13 +22,39 @@ async def test_configuration_parameter_values(
...
+    value = node_2.values["31-112-0-8-255"]
+    assert len(ack_commands_2) == 1
+    assert ack_commands_2[0] == {
+        "command": "node.set_value",
+        "nodeId": node_2.node_id,
+        "valueId": value.data,
+        "value": 190,
+        "messageId": uuid4,
+    }
+
```
raman325, line 57: Can we add an additional success test for setting this value using a valid state label?

firstof9: Sure

raman325 (approved):

test/util/test_node.py
```diff
@@ -21,13 +22,51 @@ async def test_configuration_parameter_values(
...
+    assert len(config_values_2) == 15
+
     for value in config_values.values():
         assert isinstance(value, ConfigurationValue)
 
     # Test setting a configuration parameter that has no metadata
     with pytest.raises(NotImplementedError):
         await async_set_config_parameter(node, 1, 2)
 
+    # Test setting an manual entry configuration parameter with an invalid value
```
MartinHjelmare, line 36:
```suggestion
    # Test setting a manual entry configuration parameter with an invalid value
```

test/util/test_node.py
```diff
@@ -21,13 +22,51 @@ async def test_configuration_parameter_values(
...
+    with pytest.raises(InvalidNewValue):
+        await async_set_config_parameter(node_2, "Purple", 8, 255)
+
+    # Test setting an manual entry configuration parameter with an valid value
```
MartinHjelmare, line 40:
```suggestion
    # Test setting a manual entry configuration parameter with a valid value
```

MartinHjelmare (approved):

Issue #162, opened 2021-03-11
Migrate to zwave-js 7.0.0


MartinHjelmare:
This issue lists changes that are required to support zwave-js 7.0.0.

- The changes require a new schema version, so we can't release any of these changes unless we bump the schema version.
- Of course, we need to coordinate the changes between server and client libraries. Most changes require a change in both libraries.

## Rename controller method

- [ ] Rename controller.removeNodeFromAllAssocations to controller.removeNodeFromAllAssociations.
  https://github.com/zwave-js/zwave-js-server/blob/e0d5dec1a66283f68e044d4beb87e3610a884bb6/src/lib/controller/command.ts#L17
  https://github.com/zwave-js/zwave-js-server/blob/e0d5dec1a66283f68e044d4beb87e3610a884bb6/src/lib/controller/incoming_message.ts#L99-L121
  https://github.com/zwave-js/zwave-js-server/blob/e0d5dec1a66283f68e044d4beb87e3610a884bb6/src/lib/controller/message_handler.ts#L98-L99
  https://github.com/zwave-js/zwave-js-server/blob/e0d5dec1a66283f68e044d4beb87e3610a884bb6/src/lib/controller/outgoing_message.ts#L24
  https://github.com/home-assistant-libs/zwave-js-server-python/blob/876ef9226d1a19d36aa669cb8553dcf8c2b34a6d/zwave_js_server/model/controller.py#L321-L328

## Rework node properties

- [ ] Change isFrequentListening to have the type FLiRS = false | "250ms" | "1000ms" (previously boolean) to indicate the wakeup frequency.
  https://github.com/zwave-js/zwave-js-server/blob/e0d5dec1a66283f68e044d4beb87e3610a884bb6/src/lib/state.ts#L83
  https://github.com/zwave-js/zwave-js-server/blob/e0d5dec1a66283f68e044d4beb87e3610a884bb6/src/lib/state.ts#L177
  https://github.com/home-assistant-libs/zwave-js-server-python/blob/876ef9226d1a19d36aa669cb8553dcf8c2b34a6d/zwave_js_server/model/node.py#L53
  https://github.com/home-assistant-libs/zwave-js-server-python/blob/876ef9226d1a19d36aa669cb8553dcf8c2b34a6d/zwave_js_server/model/node.py#L158-L161
- [ ] Rename maxBaudRate to maxDataRate.
[...]

PR #168, opened 2021-03-18
Handle server schema version 3

raman325:
Refer to https://github.com/home-assistant-libs/zwave-js-server-python/issues/162 for a reference of what all has been changed


zwave_js_server/model/log_config.py
```diff
@@ -42,9 +42,9 @@ def to_dict(self) -> LogConfigDataType:
     def from_dict(data: LogConfigDataType) -> "LogConfig":
         """Return LogConfig from LogConfigDataType dict."""
         return LogConfig(
-            data.get("enabled"),
-            data.get("level"),
-            data.get("logToFile"),
-            data.get("filename"),
-            data.get("forceConsole"),
+            data["enabled"],
```
MartinHjelmare, line 45: Are all keys required now? Then we can update `LogConfigDataType` to have total  True.

zwave_js_server/model/node.py
```diff
@@ -156,34 +163,44 @@ def is_listening(self) -> Optional[bool]:
...
-        """Return the max_baud_rate."""
-        return self.data.get("maxBaudRate")
+    def max_data_rate(self) -> Optional[int]:
+        """Return the max_data_rate."""
+        return self.data.get("maxDataRate")
+
+    @property
+    def supported_data_rates(self) -> Optional[List[int]]:
+        """Return the supported_data_rates."""
+        return self.data.get("supportedDataRates")
```
MartinHjelmare, line 183: Containers should always default to empty container.
```suggestion
        return self.data.get("supportedDataRates", [])
```

zwave_js_server/model/node.py
```diff
@@ -156,34 +163,44 @@ def is_listening(self) -> Optional[bool]:
...
     @property
-    def max_baud_rate(self) -> Optional[int]:
-        """Return the max_baud_rate."""
-        return self.data.get("maxBaudRate")
+    def max_data_rate(self) -> Optional[int]:
+        """Return the max_data_rate."""
+        return self.data.get("maxDataRate")
+
+    @property
+    def supported_data_rates(self) -> Optional[List[int]]:
```
MartinHjelmare, line 181:
```suggestion
    def supported_data_rates(self) -> List[int]:
```

zwave_js_server/model/notification.py
```diff
@@ -11,19 +11,48 @@
...
+    """Represent a generic notification event data dict type."""
 
     source: Literal["node"]  # required
     event: Literal["notification"]  # required
     nodeId: int  # required
-    notificationLabel: str  # required
+    ccId: int  # required
+
+
+class EntryControlNotificationArgsDataType(TypedDict, total=False):
```
MartinHjelmare, line 22: If  all keys are required we can set total to True.
```suggestion
class EntryControlNotificationArgsDataType(TypedDict, total=True):
```

zwave_js_server/model/log_config.py
raman325: I don't think they ever were optional

raman325: Actually, in the reverse (when we update the log config), we support sending partial parameters, so I don't think we can set `Total` to `False`

zwave_js_server/model/notification.py
raman325: `eventData` is optional https://zwave-js.github.io/node-zwave-js/#/api/node?id=quotnotificationquot. Will update the PR accordingly

raman325:
`eventData` in `EntryControlNotificationArgsDataType` can be a buffer. Should we parse the buffer like we do for values?

zwave_js_server/model/notification.py
```diff
@@ -11,19 +11,48 @@
 
 
 class NotificationDataType(TypedDict, total=False):
```
MartinHjelmare, line 13: Can we change here?

zwave_js_server/model/notification.py
```diff
@@ -34,11 +64,66 @@ def node_id(self) -> int:
...
+    @property
+    def data_type(self) -> int:
+        """Return data type property."""
+        return self.data["args"]["dataType"]
+
+    @property
+    def event_data(self) -> Optional[str]:
+        """Return event data property."""
+        event_data = self.data["args"].get("eventData")
+        return parse_buffer(event_data) if event_data else cast(str, event_data)
```
MartinHjelmare, line 129: Why do we cast to string if `event_data` is falsy? Looks like we're potentially casting `None` to string.

raman325: good catch

raman325:
damn, that extra line took me out of the patch requirement 🙁  I could switch it back but I think this reads better

MartinHjelmare:
We want to test the entry control CC notification anyway, right?

raman325:
> We want to test the entry control CC notification anyway, right?

Ideally yes. I can fake a payload for it but would be great to get one from an actual device

raman325:
Are we ok with proceeding with this as is and adding the Entry Control CC tests later? I'm not sure that we'll be able to get test data quickly if at all

MartinHjelmare:
What devices produce those notifications? I have a lock.

raman325:
> What devices produce those notifications? I have a lock.

I believe it is Secure Keypad Specific Device Class -> https://www.silabs.com/documents/public/miscellaneous/SDS10242-Z-Wave-Device-Class-Specification.pdf page 80

raman325:
It looks like there are three products that qualify for this:
https://products.z-wavealliance.org/products/2478?selectedFrequencyId=2
https://products.z-wavealliance.org/products/3870?selectedFrequencyId=2
https://products.z-wavealliance.org/products/3575?selectedFrequencyId=-1

raman325:
Thanks goes to @WizKid for providing the test data for the last test

MartinHjelmare (approved):
Nice!

PR #172, opened 2021-03-20
Add device class property to endpoint model

raman325:
Part of the changes for `zwave-js 7.0.0`, introduced in beta2. The `device_class` on the endpoint is the same as what we expect on the node


zwave_js_server/model/endpoint.py
```diff
@@ -14,6 +16,7 @@ class EndpointDataType(TypedDict, total=False):
     index: int  # required
     installerIcon: int
     userIcon: int
+    deviceClass: DeviceClassDataType
```
MartinHjelmare, line 19: Is this a required item?

raman325: Ohh yes thought I had added that

MartinHjelmare (approved):
Looks good!

PR #176, opened 2021-03-23
Add utility function to bulk set partial config parameters

raman325:
This has been an in demand request for some time. The new function will either accept a raw integer value or a dictionary of `{bitmask: value}` and will convert the dictionary to the raw integer value before calling the `node.set_value` command. ~~The logic all works, what I am not sure of is if we can use `node.set_value` to bulk set parameters. I asked @petro to test this for me because I don't have any nodes with partial parameters~~ EDIT: Tests confirmed that we can in fact use `node.set_value` to bulk set parameters, so this PR is good to merge

### DIFF
--- a/zwave_js_server/model/node.py
+++ b/zwave_js_server/model/node.py
@@ -325,7 +325,7 @@ class Node(EventBase):
 
         self.emit(event.type, event.data)
 
-    async def _async_send_command(
+    async def async_send_command(
         self,
         cmd: str,
         require_schema: Optional[int] = None,
@@ -367,7 +367,7 @@ class Node(EventBase):
             raise UnwriteableValue
 
         # the value object needs to be send to the server
-        result = await self._async_send_command(
+        result = await self.async_send_command(
             "set_value",
             valueId=val.data,
             value=new_value,
@@ -381,11 +381,11 @@ class Node(EventBase):
 
     async def async_refresh_info(self) -> None:
         """Send refreshInfo command to Node."""
-        await self._async_send_command("refresh_info", wait_for_result=False)
+        await self.async_send_command("refresh_info", wait_for_result=False)
 
     async def async_get_defined_value_ids(self) -> List[Value]:
         """Send getDefinedValueIDs command to Node."""
-        data = await self._async_send_command(
+        data = await self.async_send_command(
             "get_defined_value_ids", wait_for_result=True
         )
 
@@ -403,21 +403,21 @@ class Node(EventBase):
         if not isinstance(val, Value):
             val = self.values[val]
         # the value object needs to be send to the server
-        data = await self._async_send_command(
+        data = await self.async_send_command(
             "get_value_metadata", valueId=val.data, wait_for_result=True
         )
         return ValueMetadata(cast(MetaDataType, data))
 
     async def async_abort_firmware_update(self) -> None:
         """Send abortFirmwareUpdate command to Node."""
-        await self._async_send_command("abort_firmware_update", wait_for_result=False)
+        await self.async_send_command("abort_firmware_update", wait_for_result=False)
 
     async def async_poll_value(self, val: Union[Value, str]) -> None:
         """Send pollValue command to Node for given value (or value_id)."""
         # a value may be specified as value_id or the value itself
         if not isinstance(val, Value):
             val = self.values[val]
-        await self._async_send_command("poll_value", valueId=val.data, require_schema=1)
+        await self.async_send_command("poll_value", valueId=val.data, require_schema=1)
 
     def handle_wake_up(self, event: Event) -> None:
         """Process a node wake up event."""
--- a/zwave_js_server/util/node.py
+++ b/zwave_js_server/util/node.py
@@ -3,7 +3,7 @@ import json
 from typing import Dict, Optional, Union, cast
 
 from ..const import CommandClass, ConfigurationValueType
-from ..exceptions import InvalidNewValue, NotFoundError, SetValueFailed
+from ..exceptions import InvalidNewValue, NotFoundError, SetValueFailed, ValueTypeError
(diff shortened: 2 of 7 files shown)


raman325:
One thing to note is that Al has stated that he'd like to add native support for this type of action in `zwave-js`. That could eliminate or significantly reduce the amount of logic that has to live in the library. I personally think it's nice to have all of this logic in the library because we can raise relevant exceptions for the user rather than showing a generic error, but maybe that's just me

zwave_js_server/util/node.py
```diff
@@ -115,6 +74,158 @@ async def async_set_config_parameter(
...
+        # each partial parameter's metadata
+        for property_key in sorted(
+            [cast(int, value.property_key) for value in property_values], reverse=True
+        ):
+            multiplication_factor = 2 ** partial_param_bit_shift(property_key)
+            partial_value = int(remaining_value / multiplication_factor)
+            remaining_value = remaining_value % multiplication_factor
+            zwave_value = cast(
+                ConfigurationValue,
+                node.values[
```
MartinHjelmare, line 154: Do we need to guard against a missing value id here?

raman325: We shouldn't have to because we are looping through property keys that we retrieved from the node, but I just realized there's a better way to do this.

raman325:
FYI `node.set_value` does work for setting bulk values, so this PR is good to go whenever it's approved

firstof9:
Results from the test if anyone wants them.

Test device `Inovelli On/Off Red Series Switch`
Data sent:
```
Value: 51972798
Parameter: 8
```

<details><summary>Test Results</summary>
```log
{ type: 'result',
  success: true,
  messageId: 'set-parameter',
  result: { success: true } }
{ type: 'event',
  event:
   { source: 'node',
     event: 'value updated',
     nodeId: 27,
     args:
      { commandClassName: 'Configuration',
        commandClass: 112,
        property: 8,
        propertyKey: 255,
        newValue: 190,
        prevValue: 21,
        propertyName: 'LED Effect Color' } } }
{ type: 'event',
  event:
   { source: 'node',
     event: 'value updated',
     nodeId: 27,
     args:
      { commandClassName: 'Configuration',
        commandClass: 112,
        property: 8,
        propertyKey: 65280,
        newValue: 10,
        prevValue: 10,
        propertyName: 'LED Effect Brightness' } } }
{ type: 'event',
  event:
   { source: 'node',
     event: 'value updated',
     nodeId: 27,
     args:
      { commandClassName: 'Configuration',
        commandClass: 112,
        property: 8,
        propertyKey: 16711680,
        newValue: 25,
        prevValue: 10,
        propertyName: 'LED Effect Duration' } } }
{ type: 'event',
  event:
   { source: 'node',
     event: 'value updated',
     nodeId: 27,
     args:
      { commandClassName: 'Configuration',
        commandClass: 112,
        property: 8,
        propertyKey: 2130706432,
        newValue: 3,
        prevValue: 4,
        propertyName: 'LED Effect Type' } } }
```
</details>

zwave_js_server/util/node.py
```diff
@@ -115,6 +74,147 @@ async def async_set_config_parameter(
...
+    new_value: Union[int, Dict[int, Union[int, str]]],
+) -> None:
+    """Bulk set partial configuration values on this node."""
+    config_values = node.get_configuration_values()
+    property_values = [
+        value for value in config_values.values() if value.property_ == property_
+    ]
+
+    # If we can't find any values with this property, the property is wrong
+    if len(property_values) == 0:
```
MartinHjelmare, line 99:
```suggestion
    if not property_values:
```

zwave_js_server/util/node.py
```diff
@@ -115,6 +74,147 @@ async def async_set_config_parameter(
...
+
+    response = await node.async_send_command(
+        "set_value",
+        valueId={
+            "commandClass": CommandClass.CONFIGURATION.value,
+            "property": property_,
+        },
+        value=new_value,
+    )
+    if response and cast(bool, response["success"]) is False:
```
MartinHjelmare, line 164:
```suggestion
    if response and not cast(bool, response["success"]):
```

MartinHjelmare: If there's no response, is that not an error?

zwave_js_server/model/node.py
```diff
@@ -325,7 +325,7 @@ def receive_event(self, event: Event) -> None:
 
         self.emit(event.type, event.data)
 
-    async def _async_send_command(
+    async def async_send_command(
```
MartinHjelmare, line 328: Why are we changing these?

zwave_js_server/util/node.py
raman325: No. If we get no response (None) it means we called send_command_no_wait so we aren't waiting for a response

zwave_js_server/model/node.py
raman325: I changed this because I access this method in the new utility function. The method can't be private as a result

MartinHjelmare:
What do we need to change if zwave-js adds more native support for this feature?

raman325:
Not sure, because I am not sure what adding more native support means. I'm guessing that the main change would be to not calculate the raw value to send by doing the bit shift operations. All the other validation I would think we want to do in the library because we can raise more relevant exceptions to the caller

MartinHjelmare:
Ok, so there's no risk of breaking changes for the user if this would happen?

raman325:
I don't believe so. This method should work as long as the `property_key` (bitmask for config parameters) and `value` format stay the same. If there was a change to that, I would suggest we handle it as a schema bump because that's a pretty significant change IMO

MartinHjelmare (approved):
Ok, looks good!

PR #177, opened 2021-03-23
Return command status when setting a config parameter

raman325:
@cgarwood  has requested that we have a way to know whether the command failed, was accepted, or was queued when setting a config parameter. This PR adds a status to the return which can be used in HA to get the status.

This should follow #176 so we can apply this to both utility functions

### DIFF
--- a/test/util/test_node.py
+++ b/test/util/test_node.py
@@ -52,7 +52,9 @@ async def test_configuration_parameter_values(
         {"success": True},
     )
 
-    await async_set_config_parameter(node_2, 190, 8, 255)
+    zwave_value, cmd_status = await async_set_config_parameter(node_2, 190, 8, 255)
+    assert isinstance(zwave_value, ConfigurationValue)
+    assert cmd_status == CommandStatus.ACCEPTED
 
     value = node_2.values["31-112-0-8-255"]
     assert len(ack_commands_2) == 1
@@ -64,7 +66,9 @@ async def test_configuration_parameter_values(
         "messageId": uuid4,
     }
 
-    await async_set_config_parameter(node_2, "Blue", 8, 255)
+    zwave_value, cmd_status = await async_set_config_parameter(node_2, "Blue", 8, 255)
+    assert isinstance(zwave_value, ConfigurationValue)
+    assert cmd_status == CommandStatus.ACCEPTED
 
     value = node_2.values["31-112-0-8-255"]
     assert len(ack_commands_2) == 2
@@ -97,9 +101,11 @@ async def test_configuration_parameter_values(
         await async_set_config_parameter(node, 1, 1, property_key=1)
 
     # Test setting a configuration parameter by state label and property name
-    await async_set_config_parameter(
+    zwave_value, cmd_status = await async_set_config_parameter(
         node, "2.0\u00b0 F", "Temperature Reporting Threshold"
     )
+    assert isinstance(zwave_value, ConfigurationValue)
+    assert cmd_status == CommandStatus.ACCEPTED
 
     value = node.values["13-112-0-1"]
     assert len(ack_commands) == 3
@@ -119,7 +125,8 @@ async def test_bulk_set_partial_config_parameters(multisensor_6, uuid4, mock_com
         {"command": "node.set_value", "nodeId": node.node_id},
         {"success": True},
     )
-    await async_bulk_set_partial_config_parameters(node, 101, 241)
+    cmd_status = await async_bulk_set_partial_config_parameters(node, 101, 241)
+    assert cmd_status == CommandStatus.QUEUED
     assert len(ack_commands) == 1
     assert ack_commands[0] == {
         "command": "node.set_value",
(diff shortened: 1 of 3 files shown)


zwave_js_server/util/node.py
```diff
@@ -115,12 +74,160 @@ async def async_set_config_parameter(
...
+    cmd_response = await node.async_set_value(zwave_value, new_value)
+    if cmd_response is False:
         raise SetValueFailed(
             "Unable to set value, refer to "
             "https://zwave-js.github.io/node-zwave-js/#/api/node?id=setvalue for "
             "possible reasons"
         )
 
-    return zwave_value
+    status = CommandStatus.ACCEPTED if cmd_response else CommandStatus.QUEUED
```
AlCalzone, line 231: I'm not sure this does what you expect (or I don't understand python enough).
When you call zwave-js's setValue API one of those things happens:
- `false` is returned if the API call cannot be done
- A `Promise` is returned if the API call can be done (and the message was queued)

The message stays in the queue until it is its turn to be sent. This can be almost immediately or worst case take hours until the device wakes up. Once the message was sent, again two things may happen (after a short while):
- The promise resolves to `true` if the command succeeded
- The promise rejects if the command did not succeed

For the calling code, this means `await`ing the call can result in three possible situations (for `cmd_response`):
- `false` (immediately): The call could not be done
- `true` (some time in the future - may be milliseconds, may be hours): The call succeeded
- Error (some time in the future - may be milliseconds, may be hours): The call failed

zwave_js_server/util/node.py
```diff
@@ -115,12 +74,160 @@ async def async_set_config_parameter(
...
+    # into partial params
+    if len(property_values) == 1:
+        raise ValueTypeError(
+            f"Configuration parameter {property_} for node {node.node_id} does not "
+            "have partials"
+        )
+
+    # If new_value is a dictionary, we need to calculate the full value to send
+    if isinstance(new_value, dict):
+        temp_value = 0
```
AlCalzone, line 114: If you don't set **all** partials, this needs to be the currently cached (full) value for the config param.

zwave_js_server/util/node.py
raman325: We have two paths for sending a command: in one, we await the promise, in the other we don't. We determine which path to use based on the node status - if the node is asleep, we don't await the promise, if the node is any other status, we do.

When we await the promise, we expect true or false to be returned. If false, we raise an exception because something failed, if true, then in this PR we consider the `CommandStatus` to be `ACCEPTED`. When we don't await the promise, we assume that it's been queued. We have no way of verifying its future success/failure at that point. That logic is what has been implemented here.

I don't think I have seen the third situation you mentioned, the error scenario, but I don't think we account for it.

zwave_js_server/util/node.py
raman325: Right, we grab the missing partials values to add to the payload here: https://github.com/home-assistant-libs/zwave-js-server-python/pull/177/files#diff-f10e0f95086413d4ac47fadfd902b56d63ea3763366a8a8cb988cbdc7184a52cR132

zwave_js_server/util/node.py
AlCalzone: I see.

> I don't think I have seen the third situation you mentioned, the error scenario, but I don't think we account for it.

Should be somewhat easy to test. Get a switchable plug, let it be interviewed, then pull it out of the socket before you try to change its configuration. The command will be handled immediately but fail due to a missing ACK.

raman325: I just checked the server code, and we catch any errors that are thrown and return `success` as `False`, so I think we cover all three scenarios with this logic

test/util/test_node.py
```diff
@@ -171,3 +176,44 @@ async def test_bulk_set_partial_config_parameters(multisensor_6, uuid4, mock_com
...
         await async_bulk_set_partial_config_parameters(node, 252, 1)
+
+
+async def test_failures(multisensor_6, mock_command):
+    """Test setting config parameter failures."""
+    node: Node = multisensor_6
+    # We need the node to be alive so we wait for a response
+    node.handle_alive(node)
+
+    _ = mock_command(
```
MartinHjelmare, line 187:
```suggestion
    mock_command(
```

test/util/test_node.py
```diff
@@ -171,3 +176,44 @@ async def test_bulk_set_partial_config_parameters(multisensor_6, uuid4, mock_com
...
+        await async_set_config_parameter(node, 1, 101, 64)
+
+
+async def test_returned_values(multisensor_6, mock_command):
+    """Test returned values from setting config parameters."""
+    node: Node = multisensor_6
+    # We need the node to be alive so we wait for a response
+    node.handle_alive(node)
+
+    _ = mock_command(
```
MartinHjelmare, line 207:
```suggestion
    mock_command(
```

zwave_js_server/const.py
```diff
@@ -10,6 +10,14 @@
 VALUE_UNKNOWN = "unknown"
 
 
+class CommandStatus(str, Enum):
+    """Status of a command sent to zwave-js-server."""
+
+    ACCEPTED = "accepted"
+    QUEUED = "queued"
+    FAILED = "failed"
```
MartinHjelmare, line 18: We don't seem to use the `FAILED` enum. Do we still need it?

raman325: We can remove it. I was thinking it would be nice in the future if, for queued messages, we could support an optional callback in the future that returned the result once the message was processed which is why I had added this. But I will remove it for now and we can discuss that separately

MartinHjelmare (approved):
Looks good!

PR #179, opened 2021-04-03
Handle both partial and full values for config parameters

raman325:
It turns out that when we bulk set a config parameter, `zwave-js` caches the full parameter value in addition to the partials. This meant that after someone first used `bulk_set_partial_config_parameters`, it would not work on subsequent requests without restarting the server to reset the cache because the logic assumes that you either have partial values or a single full value. This is something we would have had to handle anyway if Al ends up providing separate `ZwaveValue`s for both the full value and all of the partial parameters which he has mentioned wanting to do, so this is an improvement regardless.

While I was there, we now fall back to `async_set_config_parameter` in cases where we can potentially still set the value. There's no point throwing an exception if we don't have to. We do log an info though to let the user know what happened.

We should do a patch release for this ahead of the 2021.4 release

Fixes home-assistant/core#48657

### DIFF
--- a/zwave_js_server/util/node.py
+++ b/zwave_js_server/util/node.py
@@ -113,7 +113,7 @@ async def async_bulk_set_partial_config_parameters(
             # If the new value is provided as an int, we may as well try to set it
             # using the standard utility function
             else:
-                _LOGGER.warning(
+                _LOGGER.info(
                     "Falling back to async_set_config_parameter because no partials "
                     "were found"
                 )
--- a/zwave_js_server/util/node.py
+++ b/zwave_js_server/util/node.py
@@ -118,8 +118,7 @@ async def async_bulk_set_partial_config_parameters(
                 "Falling back to async_set_config_parameter because no partials "
                 "were found"
             )
-            _, cmd_status = await async_set_config_parameter(node, new_value, property_)
-            return cmd_status
+            return (await async_set_config_parameter(node, new_value, property_))[1]
 
         # Otherwise this config parameter does not exist
         raise NotFoundError(
--- a/zwave_js_server/util/node.py
+++ b/zwave_js_server/util/node.py
@@ -135,14 +135,13 @@ async def async_bulk_set_partial_config_parameters(
             value_id = get_value_id(
                 node, CommandClass.CONFIGURATION, property_, property_key=property_key
             )
-            if value_id not in node.values:
+            if value_id not in config_values:
                 raise NotFoundError(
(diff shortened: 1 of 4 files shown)


zwave_js_server/util/node.py
```diff
@@ -92,23 +95,35 @@ async def async_bulk_set_partial_config_parameters(
...
+        if value.property_ == property_ and value.property_key is not None
     ]
 
     # If we can't find any values with this property, the property is wrong
     if not property_values:
+        # If we find a value with this property_, we know this value isn't split
+        # into partial params
+        if get_value_id(node, CommandClass.CONFIGURATION, property_) in config_values:
+            if isinstance(new_value, dict):
+                raise ValueTypeError(
```
firstof9, line 109: lint doesn't like this:
`zwave_js_server/util/node.py:108:12: R1720: Unnecessary "else" after "raise" (no-else-raise)`

zwave_js_server/util/node.py
```diff
@@ -92,23 +95,35 @@ async def async_bulk_set_partial_config_parameters(
...
+                    "does not have partials"
+                )
+            # If the new value is provided as an int, we may as well try to set it
+            # using the standard utility function
+            else:
+                _LOGGER.info(
+                    "Falling back to async_set_config_parameter because no partials "
+                    "were found"
+                )
+                return await async_set_config_parameter(node, new_value, property_)
```
firstof9, line 120: mypy is complaining about this line:
`zwave_js_server/util/node.py:120: error: Incompatible return value type (got "Tuple[ConfigurationValue, CommandStatus]", expected "CommandStatus")`

raman325: good catch thanks

firstof9:
linter once again: `zwave_js_server/util/node.py:90:0: R0914: Too many local variables (16/15) (too-many-locals)`

zwave_js_server/util/node.py
```diff
@@ -92,23 +95,36 @@ async def async_bulk_set_partial_config_parameters(
     """Bulk set partial configuration values on this node."""
     config_values = node.get_configuration_values()
     property_values = [
-        value for value in config_values.values() if value.property_ == property_
+        value
+        for value in config_values.values()
+        if value.property_ == property_ and value.property_key is not None
     ]
 
     # If we can't find any values with this property, the property is wrong
```
MartinHjelmare, line 103: Should we update this comment?

MartinHjelmare (approved):
Looks good!

PR #184, opened 2021-04-06
Support new interview events and add logic to track interview stage

raman325:
Follows https://github.com/zwave-js/zwave-js-server/pull/181


raman325:
It occurs to me that there is an `interviewStage` parameter that gets sent to us in the node state dump. Right now we store the enum value, but per conversation with Al, we shouldn't rely on the enum value (int) to be stable. So perhaps the best way to do this is to provide the enum name as `interviewStage` instead and then update that when we handle these events.

But if we want to do that, we will have to introduce a new schema which may not be worth it for this single change. Thoughts?

MartinHjelmare:
We shouldn't rely on things that aren't stable in zwave-js. That will break sooner or later. I think a schema bump is ok, but we can wait with the bump until closer to the next Core release and maybe more things will have changed requiring a new schema by then.

raman325:
> We shouldn't rely on things that aren't stable in zwave-js. That will break sooner or later. I think a schema bump is ok, but we can wait with the bump until closer to the next Core release and maybe more things will have changed requiring a new schema by then.

The approach of waiting is fine, but with the current approach, nothing would break because we aren't using `interviewStage` in this logic, we're storing the name of the stage as a string so if the stage names or enums change, it doesn't affect the logic.

MartinHjelmare:
Won't potential breakage depend on how we depend on the stageName string?

raman325:
I don't think we should introduce any dependencies on the stageName string. `interview started` and `interview completed` will affect the ready flag which is what we should rely on. The `is_being_interviewed` property I added is dependent on the ready property and a status that is internally set when an interview fails, and is thus safe to use.

The only utility of the stageName string IMO is to display the stage in the UI for the user to track the progress.

MartinHjelmare:
Yeah. It's important that we make sure the GUI doesn't introduce any dependencies on this string then. Eg translation or extra description.

raman325:
upstream has been merged so this is ready for review

zwave_js_server/const.py
```diff
@@ -3,12 +3,14 @@
 from typing import Dict, List
 
 # minimal server schema version we can handle
-MIN_SERVER_SCHEMA_VERSION = 3
+MIN_SERVER_SCHEMA_VERSION = 4
```
MartinHjelmare, line 6: Do we need to release a new version of the server before we merge here?

zwave_js_server/model/node.py
```diff
@@ -82,7 +82,7 @@ class NodeDataType(TypedDict, total=False):
     individualEndpointCount: int
     aggregatedEndpointCount: int
     interviewAttempts: int
-    interviewStage: int
+    interviewStage: Optional[Union[int, str]]
```
MartinHjelmare, line 85: Can this be `None` or is it just potentially missing in the dict?

zwave_js_server/model/node.py
```diff
@@ -291,10 +291,15 @@ def interview_attempts(self) -> Optional[int]:
         return self.data.get("interviewAttempts")
 
     @property
-    def interview_stage(self) -> Optional[int]:
+    def interview_stage(self) -> Optional[Union[int, str]]:
         """Return the interview_stage."""
         return self.data.get("interviewStage")
 
+    @property
+    def is_being_interviewed(self) -> bool:
```
MartinHjelmare, line 299: Maybe rename to `in_interview`?

zwave_js_server/const.py
raman325: We need to release a new version of the server before we **release**. We can hold off on merge until we do that, but every PR thats open right now, plus one I haven't submitted yet to add commands related to statistics collection, is dependent on that release

zwave_js_server/model/node.py
raman325: I'm not sure why I added Optional here.. Will check the server code and update accordingly

raman325: oh the reason is because I set it to None when the interview restarts since this indicates the last interview stage completed: https://github.com/home-assistant-libs/zwave-js-server-python/pull/184/files#diff-2236aec645937ebf73f92b60e09e076889c45c19d46c86f4ecb47d589d276c47R451

MartinHjelmare (approved):
Looks good!

zwave_js_server/const.py
MartinHjelmare: I'll let you coordinate the releases and merges. :+1:

raman325:
are you aware of any outstanding issues with schema 3? The next release for `zwave-js-server` will add schema 4 support, so I think this is safe to merge unless you foresee any hotfixes needed for our current iteration of the integration.

Separately, I am wondering if we should consider maintaining branches as we switch schemas, so that we can move forward with developing for new schemas but maintain a path to release hot fixes for old ones

MartinHjelmare:
I'm not aware of anything specific to schema 3, as far as I have deducted.

Most issues are device specific. Here are some issues that may not be device specific:
- Firmware version not updated in device info.
- Lack of USB path options in config flow.
- Lack of options flow to reconfigure the server (instead of going through the config flow again).
- Missing feedback from some light and climate devices after command.
- Device configuration don't stick.
- Climate temperature unit reporting mismatch.

Sounds good to use separate target branches for the next schema to be released. :+1:

PR #188, opened 2021-04-08
Support property names in bulk_set_partial_config_parameters

raman325:
Right now if you provide a dictionary of partial parameter values to set for `bulk_set_partial_config_parameters`, it's assumed that the keys are an integer representation of a bitmask. For further convenience, we are adding the ability for users to specify the property name of the partial parameter as the key instead.

Note that because most of the logic lives in the library, the only change that is needed on the HA side is a bump in the lib version and a change to the service schema to allow strings to be passed

### DIFF
--- a/zwave_js_server/util/node.py
+++ b/zwave_js_server/util/node.py
@@ -81,7 +81,7 @@ async def async_set_config_parameter(
 async def async_bulk_set_partial_config_parameters(
     node: Node,
     property_: int,
-    new_value: Union[int, Dict[int, Union[int, str]]],
+    new_value: Union[int, Dict[Union[int, str], Union[int, str]]],
 ) -> CommandStatus:
     """Bulk set partial configuration values on this node."""
     config_values = node.get_configuration_values()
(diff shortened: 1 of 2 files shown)


zwave_js_server/util/node.py
```diff
@@ -222,33 +222,55 @@ def _get_int_from_partials_dict(
...
+                )
+            except StopIteration:
+                raise NotFoundError(
+                    f"Partial parameter with label '{property_key_or_name}'"
+                    f"not found for parameter {property_}"
+                ) from None
+
+        provided_partial_values.append(zwave_value)
+        partial_value = _validate_and_transform_new_value(zwave_value, partial_value)
+        int_value += partial_value << partial_param_bit_shift(zwave_value.property_key)
```
firstof9, line 265: mypy: Argument 1 to "partial_param_bit_shift" has incompatible type "Union[str, int, None]"; expected "int"

MartinHjelmare (approved):
Looks good!

PR #191, opened 2021-04-18
Add support to receive log messages

raman325:
We're adding upstream support to forward logs as driver events (https://github.com/zwave-js/zwave-js-server/pull/195). This PR adds support in the library.


zwave_js_server/model/log_message.py
```diff
@@ -0,0 +1,81 @@
...
+            return self.data[field_name].splitlines()
+
+        # We will assume each item in the array is on a separate line so we can
+        # remove trailing line breaks
+        return [message.rstrip("\n") for message in self.data[field_name]]
+
+    @property
+    def message(self) -> List[str]:
+        """Return message."""
+        self.data["message"].rstrip("\n")
```
MartinHjelmare, line 40: This doesn't change the value in `self.data`.

zwave_js_server/model/driver.py
```diff
@@ -39,19 +40,40 @@ def receive_event(self, event: Event) -> None:
...
         self.emit(event.type, event.data)
 
+    def handle_logging(self, event: Event) -> None:
+        """Process a driver logging event."""
+        event.data["log_message"] = LogMessage(cast(LogMessageDataType, event.data))
+
     def handle_all_nodes_ready(self, event: Event) -> None:
         """Process a driver all nodes ready event."""
 
+    async def async_send_command(
```
MartinHjelmare, line 50: Should this be a private method?

raman325: should we do the same for the node method?

MartinHjelmare: I haven't reviewed our use of it. If we don't use it in the integration, I think we can make it private.

raman325:
The upstream PR has been merged so this is ready for review

zwave_js_server/model/driver.py
```diff
@@ -65,6 +70,14 @@ async def async_get_log_config(self) -> LogConfig:
         result = await self._async_send_command("get_log_config", require_schema=4)
         return LogConfig.from_dict(result["config"])
 
+    async def async_start_listening_logs(self) -> None:
```
MartinHjelmare, line 73: Side note: We've talked about removing the `async_` prefix from all names, since everything should run in async context. Something for the future.

zwave_js_server/model/log_message.py
```diff
@@ -0,0 +1,88 @@
...
+        """Return fully formatted message."""
+        return self._process_message("formattedMessage")
+
+    @property
+    def direction(self) -> str:
+        """Return direction."""
+        return self.data["direction"]
+
+    @property
+    def level(self) -> Optional[str]:
```
MartinHjelmare, line 56: This doesn't match with the data dict typing.

raman325: thanks, errant copy/paste

MartinHjelmare (approved):
Looks good!

PR #196, opened 2021-04-19
Add missed commit for driver.async_enable_statistics

raman325:
`driver.enable_statistics` takes two inputs that were missed in the last PR

### DIFF
--- a/test/model/test_driver.py
+++ b/test/model/test_driver.py
@@ -119,18 +119,18 @@ async def test_statistics(driver, uuid4, mock_command):
     ack_commands = mock_command(
         {
             "command": "driver.enable_statistics",
-            "applicationName": "test",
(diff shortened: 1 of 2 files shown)


test/model/test_driver.py
```diff
@@ -117,14 +117,20 @@ async def test_statistics(driver, uuid4, mock_command):
     """Test statistics commands."""
     # Test that enable_statistics command is sent
     ack_commands = mock_command(
-        {"command": "driver.enable_statistics"},
+        {
+            "command": "driver.enable_statistics",
+            "applicationName": "test",
+            "applicationVersion": "test",
```
MartinHjelmare, line 123: Use two different test values so we know that they go to the right parameter.

MartinHjelmare (approved):
Thanks!

PR #198, opened 2021-04-19
Auto bump version on release

firstof9:
This should automatically bump the version number on release.

### DIFF
--- a/.github/workflows/publish-to-pypi.yml
+++ b/.github/workflows/publish-to-pypi.yml
@@ -10,6 +10,12 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@master
+      - name: Get Version
+        id: get_version
+        run: echo ::set-output name=VERSION::${GITHUB_REF/refs\/tags\//}
+      - name: "Set version numbmer"
+        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-to-pypi.yml
```diff
@@ -10,6 +10,21 @@ jobs:
...
+        run: |
+          sed -i '/VERSION = /c\VERSION = "${{ steps.get_version.outputs.VERSION }}"' ${{ github.workspace }}/setup.py
+      - name: Commit files
+        run: |
+          if [ $(git diff HEAD | wc -l) -gt 0 ]
+          then
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+          git config user.name "GitHub Actions"
+          git commit -a --amend --no-edit || true
+          git push --force || true
```
MartinHjelmare, line 26: I don't think it's a good idea to amend and force push to main branch. We want to bump version before the release is triggered.

If we want to automate the release further I suggest we instead auto release if the action detects an increase of version number in a push. See eg the hyper modern cookiecutter repo for an example of this workflow.

firstof9: No problem.

PR #205, opened 2021-05-04
Forward ZJS errors

balloob:
Python PR to match https://github.com/zwave-js/zwave-js-server/pull/210

Will raise Z-Wave specific errors.


MartinHjelmare (approved):
Looks good!

Looks like a new server with this feature is compatible with older clients, so we don't need a protocol change. Older clients will just not have as much info about the error.

balloob:
Correct.

PR #206, opened 2021-05-09
Improve device config model

raman325:
The existing model was pretty limited in that it returned a bunch of dictionaries that the downstream code had to parse. Here we've extended the model further so that it's easier to access all of the properties. I also added a model for the `associations` but thus far haven't been able to find test data to test it. I can revert it to a dictionary if desired until we get the test data.

I also noticed that we created the device config object every time the property was accessed, but the device config won't change until the next restart so we can safely create it during initialization.

### DIFF
--- a/zwave_js_server/model/device_config.py
+++ b/zwave_js_server/model/device_config.py
@@ -56,55 +56,6 @@ class DeviceFirmwareVersionRange:
         return self.data.get("max")
 
 
-class DeviceConditionalAssociationDataType(TypedDict, total=False):
-    """Represent a device conditional association data dict type."""
-
-    groupId: int
-    label: str
-    description: str
-    maxNodes: int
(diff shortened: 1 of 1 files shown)


raman325:
I can't figure out how to resolve the following two mypy errors:
```
zwave_js_server/model/device_config.py:180: error: Argument 1 to "DeviceFirmwareVersionRange" has incompatible type "Union[DeviceFirmwareVersionRange, Dict[<nothing>, <nothing>]]"; expected "DeviceFirmwareVersionRangeDataType"
zwave_js_server/model/device_config.py:183: error: Argument 1 to "DeviceConditionalAssociation" has incompatible type "Dict[Any, Any]"; expected "DeviceConditionalAssociationDataType"
```

MartinHjelmare (approved):
Looks good!

PR #207, opened 2021-05-10
Handle firmware update events better

raman325:
Right now we don't handle firmware update events. This PR adds firmware update event models that we add to the event. This is in preparation for supporting initiating firmware updates from the lib


MartinHjelmare (approved):
Looks good!

PR #208, opened 2021-05-10
Add support for node.begin_firmware_update

raman325:
Added support for one new command:
- `node.begin_firmware_update`

This command allows the caller to begin a firmware update with both a known and unknown file format. We will likely use the unknown file format approach in HA.

~~This will be in draft until https://github.com/zwave-js/zwave-js-server/pull/213 gets merged~~

Because of the size of the files that we will be sending, we will be using a separate websocket connection from the main one to send the request

### DIFF
--- a/zwave_js_server/firmware.py
+++ b/zwave_js_server/firmware.py
@@ -1,0 +1,71 @@
+"""Dump helper."""
+from typing import Any, Dict
+import uuid
+
+import aiohttp
+
+from .const import MAX_SERVER_SCHEMA_VERSION
+from .model.node import Node
+from .util.helpers import convert_bytes_to_base64
+
+
+async def begin_firmware_update_guess_format(
+    url: str, node: Node, filename: str, file: bytes, session: aiohttp.ClientSession
+) -> Dict[str, Any]:
+    """Send beginFirmwareUpdate command to Node (format to be guessed)."""
+    client = await session.ws_connect(url)
+    # Version info
+    await client.receive_json()
+    await client.send_json(
+        {
+            "command": "set_api_schema",
+            "messageId": "api-schema-id",
+            "schemaVersion": MAX_SERVER_SCHEMA_VERSION,
+        }
+    )
+    # set_api_schema response
+    await client.receive_json()
+
+    await client.send_json(
+        {
+            "command": "node.begin_firmware_update_guess_format",
+            "nodeId": node.node_id,
+            "firmwareFilename": filename,
+            "firmwareFile": convert_bytes_to_base64(file),
+            "messageId": uuid.uuid4().hex,
+        }
+    )
+    resp = await client.receive_json()
+    await client.close()
+    return resp
+
+
+async def begin_firmware_update_known_format(
+    url: str, node: Node, file_format: str, file: bytes, session: aiohttp.ClientSession
+) -> Dict[str, Any]:
+    """Send beginFirmwareUpdate command to Node (format is known)."""
+    client = await session.ws_connect(url)
+    # Version info
+    await client.receive_json()
+    await client.send_json(
+        {
+            "command": "set_api_schema",
+            "messageId": "api-schema-id",
+            "schemaVersion": MAX_SERVER_SCHEMA_VERSION,
+        }
+    )
+    # set_api_schema response
+    await client.receive_json()
+
+    await client.send_json(
+        {
+            "command": "node.begin_firmware_update_known_format",
+            "nodeId": node.node_id,
+            "firmwareFileFormat": file_format,
+            "firmwareFile": convert_bytes_to_base64(file),
+            "messageId": uuid.uuid4().hex,
+        }
+    )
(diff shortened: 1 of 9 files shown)


raman325:
@MartinHjelmare can you help me figure out why my tests are failing? The response I am receiving is the network state, not the response that I've mocked. I suspect it has something to do with the fixtures, because we should only receive the network state from the server when we issue the `start_listening` command which I don't do here.

zwave_js_server/firmware.py
```diff
@@ -0,0 +1,70 @@
...
+from .const import MAX_SERVER_SCHEMA_VERSION
+from .model.node import Node
+from .util.helpers import convert_bytes_to_base64
+
+
+async def begin_firmware_update_guess_format(
+    url: str, node: Node, filename: str, file: bytes, session: aiohttp.ClientSession
+) -> Any:
+    """Send beginFirmwareUpdate command to Node (file format to be guessed)."""
+    client = await session.ws_connect(url)
```
MartinHjelmare, line 15: Why are we not using the client class? Eg like we do in the main module.

raman325: I made this change based on this comment: https://github.com/zwave-js/zwave-js-server/pull/213#discussion_r629456314

This is easy enough to revert if we feel it's unnecessary

CC @balloob

MartinHjelmare: What part of that comment do you mean? My point was that we could still use the Client class when opening a new connection, I think.

MartinHjelmare: Not, sure. I haven't looked at the details recently. There might be a good reason.

raman325: Probably because we wanted to catch the version info as part of the dump.

raman325:
I also made schema version configurable. This gives us flexibility in the future if we need to update the lib version in HA but the schema version has changed (we can then set the right schema as part of the update in the HA logic)

zwave_js_server/firmware.py
```diff
@@ -0,0 +1,35 @@
...
+        "nodeId": node.node_id,
+        "firmwareFilename": filename,
+        "firmwareFile": convert_bytes_to_base64(file),
+    }
+    if file_format is not None:
+        cmd["firmwareFileFormat"] = file_format
+
+    resp = await client.async_send_command(cmd, require_schema=5)
+    await client.disconnect()
+    return resp
```
MartinHjelmare, line 35: What is the response? How will it be used?

zwave_js_server/firmware.py
```diff
@@ -0,0 +1,35 @@
...
+from .util.helpers import convert_bytes_to_base64
+
+
+async def begin_firmware_update(
+    url: str,
+    node: Node,
+    filename: str,
+    file: bytes,
+    session: aiohttp.ClientSession,
+    file_format: str = None,
```
MartinHjelmare, line 17:
```suggestion
    file_format: Optional[str] = None,
```

zwave_js_server/firmware.py
```diff
@@ -0,0 +1,35 @@
+"""Firmware update helper."""
+from typing import Any
```
MartinHjelmare, line 2:
```suggestion
from typing import Any, Optional
```

zwave_js_server/firmware.py
raman325: If everything is successful, the response is an empty dictionary. There are cases where one of the underlying methods will fail, like if the file format can't be guessed, the firmware can't be extracted, or the command to begin the update fails. In those cases a `ZwaveError` should be thrown which we should catch and return as the response

raman325: Here's an example of what was returned when I sent thee server an invalid payload for thee file:
```
{
  type: 'result',
  success: false,
  messageId: 'update-firmware',
  errorCode: 'zwave_error',
  zwaveErrorCode: 1506,
  zwaveErrorMessage: 'Could not detect firmware format'
}
```

MartinHjelmare: We should decide what type of return value we want and ehat errors to raise. I don't think we should return the response directly.

raman325: I don't think this is specific to this feature. Maybe we should have a generic exception that is raised when a `zwave_error` is received?

MartinHjelmare: This is how we do it in the client.

https://github.com/home-assistant-libs/zwave-js-server-python/blob/aaa18a6fdb942de5aa2e9670d7789ef725242342/zwave_js_server/client.py#L269-L271

raman325: Do we need to do anything then? I think the failure scenario I imagined is already handled

MartinHjelmare: Yes, right, it's already handled. If we don't have any use for the response on non error, we don't need to return anything I think.

MartinHjelmare (approved):
Looks good!

PR #217, opened 2021-05-22
Add schema 5 support

raman325:
Upstream PR: https://github.com/zwave-js/zwave-js-server/pull/228

We removed `neighbors` from the node, added `deviceDatabaseUrl`, and added a new controller command to get the neighbors


MartinHjelmare (dismissed):
Thanks!

raman325:
There was one additional command added: https://github.com/zwave-js/zwave-js-server/pull/229

I added the `or {}` and a default value for the key to get around mypy

MartinHjelmare (approved):

PR #218, opened 2021-05-24
Improve handling of heal network events

raman325:
In this PR we parse the event data and return a corresponding dictionary that includes `Node`s mapped to a `HealNodeStatus` enum.

Note that there is an upstream bug that needs to be resolved so that we can properly see the payload data (right now we get an empty dictionary): https://github.com/zwave-js/zwave-js-server/pull/233. This PR can be merged ahead of that getting merged because it will still work with the current payload

### DIFF
--- a/test/model/test_controller.py
+++ b/test/model/test_controller.py
@@ -1,5 +1,7 @@
 """Test the controller model."""
 import json
+from zwave_js_server.const import HealNodeStatus
+from zwave_js_server.event import Event
 from zwave_js_server.model import association as association_pkg
 from zwave_js_server.model import controller as controller_pkg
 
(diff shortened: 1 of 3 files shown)


MartinHjelmare (approved):
Looks good!

raman325:
@MartinHjelmare started working on the HA WS API side of this and realized that it's simpler to just pass on the unparsed event data since that's what the frontend would need. Do we still want to keep this PR? Generally speaking I think it's a better way to model the data so I think we keep it, but wanted to mention it in case you disagree.

MartinHjelmare:
If we don't need it, we can wait to add it. You should decide, I think.

raman325:
OK, I think we can hold off for now. I do think it's still good to add for consistency sake but we can reevaluate later

PR #220, opened 2021-05-24
Add node.ping command

raman325:
Add new `node.ping` command from https://github.com/zwave-js/zwave-js-server/pull/229

I added the `or {}` and a default value for the key to get around mypy


MartinHjelmare (approved):
Looks good!

PR #222, opened 2021-05-25
Handle log config update events and store log config state

raman325:
Now that the server will send events whenever the log config is updated, we can get the log config initially and maintain its state within the lib, updating it every time we receive an event.

One thing to note is that the server will not send a `log config updated` message to the client that called the `driver.update_log_config` command, so it is up to that client to update its state when it calls that command. We handle that here.


raman325:
Let's hold off on merging this until we make a decision on https://github.com/zwave-js/zwave-js-server/pull/243

Merging that would reduce some of the complexity introduced here to `Driver.async_update_log_config`

zwave_js_server/model/driver.py
```diff
@@ -61,9 +70,13 @@ async def _async_send_command(
 
     async def async_update_log_config(self, log_config: LogConfig) -> None:
         """Update log config for driver."""
+        log_config_dict = log_config.to_dict()
         await self._async_send_command(
-            "update_log_config", config=log_config.to_dict(), require_schema=4
+            "update_log_config", config=log_config_dict, require_schema=4
         )
+        new_log_config_dict = self.log_config.to_dict()
+        new_log_config_dict.update(log_config_dict)
```
MartinHjelmare, line 78: I was confused the first time I read this line, since we're copying from the old log config and calling the variable new log config. This makes it easier to read, I think:
```suggestion
        new_log_config_dict = {**self.log_config.to_dict(), **log_config_dict}
```

But why do we need to copy the existing log config at all? Isn't the passed in new config complete?

raman325: not necessarily, it can be a partial. In any case, the upstream PR was accepted and all clients will receive the event when the log config is updated so we can remove this logic

MartinHjelmare (approved):
There's a merge conflict.

PR #223, opened 2021-05-25
Add utility functions to send multicast commands

raman325:
These multicast commands don't really fit into our existing model, so I thought they made more sense as utility commands. We can either send a broadcast to all nodes or a multicast to a subset of nodes, and that is controlled by whether the node_id's are provided or not.

### DIFF
--- a/test/util/test_multicast.py
+++ b/test/util/test_multicast.py
@@ -19,7 +19,12 @@ async def test_endpoint_get_cc_version_multicast(
         {"version": 1},
     )
 
-    assert await async_multicast_endpoint_get_cc_version(client, 1, CommandClass.ALARM, [node1, node2]) == 1
+    assert (
+        await async_multicast_endpoint_get_cc_version(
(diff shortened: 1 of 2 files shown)


test/util/test_multicast.py
```diff
@@ -0,0 +1,106 @@
...
+
+
+async def test_endpoint_supports_cc(client, uuid4, mock_command):
+    """Test broadcast_node.supports_cc command."""
+    ack_commands = mock_command(
+        {"command": "broadcast_node.supports_cc"},
+        {"supported": True},
+    )
+
+    assert await async_multicast_endpoint_supports_cc(client, 1, CommandClass.ALARM)
```
MartinHjelmare, line 65: We should test passing a list of nodes for all of the commands that support that.

raman325: done

MartinHjelmare (approved):
Great!

PR #224, opened 2021-05-26
Add config update commands

raman325:
https://github.com/zwave-js/zwave-js-server/pull/246
https://github.com/zwave-js/zwave-js-server/pull/245

Going to try for one more server, addon, and library release ahead of beta so we are unblocked to work on the API


MartinHjelmare (approved):
Looks good!

PR #225, opened 2021-05-26
Track whether heal network is active

raman325:
As discussed with @balloob we will track the state of `isHealNetworkActive` using the initial state of the controller and the `heal network` events that come in


zwave_js_server/model/controller.py
```diff
@@ -40,6 +41,7 @@ def __init__(self, client: "Client", state: dict) -> None:
         self.client = client
         self.data: ControllerDataType = state["controller"]
         self.nodes: Dict[int, Node] = {}
+        self._is_heal_network_active = self.data["isHealNetworkActive"]
```
balloob, line 44: This seems unused ?

balloob (approved):

zwave_js_server/model/controller.py
balloob:
```suggestion
```

raman325: thanks, left it behind

PR #229, opened 2021-06-01
Fix KeyError: 'isHealNetworkActive'

kpine:
Accessing the controller `is_heal_network_active` property results in a KeyError for node-zwave-js versions prior to 7.6.0.

### DIFF
--- a/zwave_js_server/model/controller.py
+++ b/zwave_js_server/model/controller.py
@@ -145,9 +145,9 @@ class Controller(EventBase):
         return self.data.get("supportsTimers")
 
     @property
-    def is_heal_network_active(self) -> bool:
+    def is_heal_network_active(self) -> Optional[bool]:
         """Return is_heal_network_active."""
-        return self.data["isHealNetworkActive"]
+        return self.data.get("isHealNetworkActive")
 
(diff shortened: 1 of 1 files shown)


kpine:
In HA 2021.6 beta this results in a blank integration configuration page.

MartinHjelmare:
We have bumped the minimum server schema version to 5 in the library and will bump the library in Home Assistant, so users will need to install zwave-js-server 1.7.0 and zwave-js 7.6.0+ for core-2021.6.0.

That should fix this, right?

kpine:
Schema 5 was added in 1.6.0.

MartinHjelmare:
Ok, we should probably have bumped schema version in 1.7.0 to 6. Now we can't use the schema version to force a specific service version.

MartinHjelmare (approved):

kpine:
Either way, accessing the dict keys with `get()` is agnostic to any server schema changes. All of the other controller properties are setup this way, this particular one is the outlier.

MartinHjelmare:
@raman325 I planning to merge this and then bump the library.

If we increase the server schema version to 6 that should fix things too. We can think about following up with that.

PR #230, opened 2021-06-03
Change the order of messages sent to the server on connection

raman325:
Currently, on connection, we send `start_listening` before `driver.get_log_config`. Per https://github.com/home-assistant/core/issues/51334 and https://github.com/home-assistant/core/issues/51384 this can result in events getting sent to the client before the `driver.get_log_config` message response is received which breaks the order of operations in `Client.listen()`. So we will simply switch the order of messages sent so that `start_listening` is the last message to be sent.

~~As part of this change, I also added the `driver.get_log_config` message to the dump. Admittedly I did this because it was easier to adjust the fixtures when the `dump_msgs` function sent the same messages in the same order as `Client.listen()`, but I also don't see any harm in adding it. With that being said, I can adjust this if needed (we'd basically have to duplicate the `ws_client` and `client_session` fixtures specifically for the dump function).~~

Fixes https://github.com/home-assistant/core/issues/51384 and fixes https://github.com/home-assistant/core/issues/51334

### DIFF
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -83,6 +83,14 @@ def client_session_fixture(ws_client):
     return client_session
 
 
+@pytest.fixture(name="dump_client_session")
+def dump_client_session_fixture(dump_ws_client):
+    """Mock an aiohttp client session to test dumps."""
+    dump_client_session = AsyncMock(spec_set=ClientSession)
+    dump_client_session.ws_connect.side_effect = AsyncMock(return_value=dump_ws_client)
+    return dump_client_session
+
+
 @pytest.fixture(name="inovelli_switch_state", scope="session")
 def inovelli_switch_state_fixture():
     """Load the bad string meta data node state fixture data."""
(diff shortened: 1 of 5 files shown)


MartinHjelmare:
Do we have any use for the log config in the dump?

If we need to, we can override package pytest fixtures by defining the same fixture in the test module.

raman325:
It's never come up as a need. May have been more useful before we added the logging UI but perhaps not needed. I will go with the approach you suggested.

raman325:
we need different fixtures in both `test_dump` and in `test_main`. Is it better to create those fixtures locally in each of those test modules or to create a new pair of fixtures in `conftest` specifically for these dump tests?

raman325:
I am getting the following failure in the `test_dump_state` test in `test_main`:
E           aiohttp.client_exceptions.ClientConnectorError: Cannot connect to host test.org:3000 ssl:default [nodename nor servname provided, or not known]

Seems like the new fixture isn't intercepting the attempt to connect. Any ideas how I can resolve that?

MartinHjelmare:
We have overridden the `client_session` fixture locally in the `test_main.py` module. That has the correct way of patching the async context manager.

raman325:
missed that, thanks!

MartinHjelmare (approved):
Looks good!

PR #246, opened 2021-07-07
Add controller/node statistics to model

raman325:
See https://github.com/zwave-js/zwave-js-server/pull/296 

We will now be receiving an initial state of statistics for the node and controller and will receive regular updates to those statistics from the `statistics updated` event

### DIFF
--- a/test/model/test_controller.py
+++ b/test/model/test_controller.py
@@ -106,16 +106,17 @@ def test_from_state():
     assert ctrl.suc_node_id == 1
     assert ctrl.supports_timers is False
     assert ctrl.is_heal_network_active is False
+    stats = ctrl.statistics
     assert (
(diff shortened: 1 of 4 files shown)


MartinHjelmare:
There's a merge conflict.

MartinHjelmare (approved):
This is backwards compatible with servers that don't support the new statistics updated event yet, right?

MartinHjelmare:
We should remember to require schema 7 as minimum version here in the library when we want to start using this feature in the integration.

raman325:
> This is backwards compatible with servers that don't support the new statistics updated event yet, right?

yes it is which is why I didn't bump schemas here although I should have set max to 7. The other PR for the new endpoint commands does require 7 so I will bump the minimum in that one before merging

PR #247, opened 2021-07-10
Handle value updated and added the same way

raman325:
See https://github.com/home-assistant/core/issues/52791 and https://discord.com/channels/330944238910963714/800356888827002880/863475611493466152 for context. We've decided from the conversation that value added and value updated events should be handled the same way.

The problem: There are cases where we retrieve a ZwaveValue that doesn't itself have a value, and this specifically seems to happen with the Notification CC. When zwave-js receives the first value for the ZwaveValue after initialization where we've already added the value to our model in the lib, it issues a `value added` event instead of a `value updated` like we would expect. This PR handles that edge case while leaving the `value updated` event effectively as is.

Please merge this before we merge the other PRs so that we can do a patch release for this

### DIFF
--- a/test/model/test_node.py
+++ b/test/model/test_node.py
@@ -771,6 +771,9 @@ async def test_firmware_events(wallmote_central_scene: Node):
 async def test_value_added_value_exists(climate_radio_thermostat_ct100_plus):
     """Test value added event when value exists."""
     node: Node = climate_radio_thermostat_ct100_plus
+    value_id = f"{node.node_id}-128-1-isHigh"
+    value = node.values.get(value_id)
+    assert value
     event = Event(
         "value added",
         {
@@ -794,7 +797,8 @@ async def test_value_added_value_exists(climate_radio_thermostat_ct100_plus):
         },
     )
     node.receive_event(event)
-    assert f"{node.node_id}-128-1-isHigh" in node.values
+    assert value_id in node.values
+    assert value is node.values[value_id]
 
 
 async def test_value_added_new_value(climate_radio_thermostat_ct100_plus):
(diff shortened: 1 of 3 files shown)


kpine:
I tested this manually with a door sensor, looks good.

kpine:
To further explain, the zwave_js integration in HA is caching references to the node's ZWaveValues during discovery. The previous implementation of the "value added" handler was invalidating those reference, meaning the value used to represent the entity state in the platform code was no longer referring to the value being updated in this library.

test/model/test_node.py
```diff
@@ -766,3 +766,32 @@ async def test_firmware_events(wallmote_central_scene: Node):
...
+                    "readable": True,
+                    "writeable": False,
+                    "label": "Low battery level",
+                },
+                "value": True,
+            },
+        },
+    )
+    node.receive_event(event)
+    assert f"{node.node_id}-128-1-isHigh" in node.values
```
kpine, line 797: I don't think this fixture sets up the test case properly. Looks like the Low Battery level already has a value in the fixture data. What you really want is a Node value with a missing `value`, and then assert that the Value instance is the same before and after the event. Do you need new test data for that, or could you just delete the value from the fixture data?

I could provide a dump with the missing value, and then the "value added" message if needed.

raman325: I faked it.

kpine (approved):
LGTM

test/model/test_node.py
```diff
@@ -766,3 +766,65 @@ async def test_firmware_events(wallmote_central_scene: Node):
...
+                    "writeable": False,
+                    "label": "High battery level",
+                },
+                "value": True,
+            },
+        },
+    )
+    node.receive_event(event)
+    assert value_id in node.values
+    assert value is node.values[value_id]
```
MartinHjelmare, line 801:
```suggestion
    assert node.values[value_id] is value
```

MartinHjelmare (approved):

PR #248, opened 2021-07-10
Bump lib to 0.27.1

raman325:
We need to do this after merging https://github.com/home-assistant-libs/zwave-js-server-python/pull/247 because that fixes a bug with notification CC values


kpine:
Consider including #249 if you think I've correctly identified and properly fixed the issue.

MartinHjelmare:
Please mark PRs that depend on other changes as draft.

MartinHjelmare:
I think we can include #249 in this patch release. I'll let @raman325 decide.

MartinHjelmare (approved):

PR #249, opened 2021-07-10
Update node's `device_info` property during a `ready` event

kpine:
When a node handles a `ready` event from the server, it doesn't update its internal device_info cache. This means access to the `node.device_info` property by clients later on have stale data. This change updates the event handler to reset the internal device_config attribute based on the new event data.

I noticed this when adding a node to HA, the manufacturing and product info were all missing, and the name was `Node XX` (default when product info is missing). Upon inspection, the integration was passing null data (derived from the Node object) to the device registry. I assume this was a regression from a previous change.

### DIFF
--- a/zwave_js_server/model/node.py
+++ b/zwave_js_server/model/node.py
@@ -504,6 +504,9 @@ class Node(EventBase):
         """Process a node ready event."""
         # the event contains a full dump of the node
         self.data.update(event.data["nodeState"])
+        # update device config
+        if new_device_config := self.data.get("deviceConfig"):
+            self._device_config = DeviceConfig(new_device_config)
         # update/add values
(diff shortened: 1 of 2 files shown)


zwave_js_server/model/node.py
```diff
@@ -504,6 +504,9 @@ def handle_ready(self, event: Event) -> None:
         """Process a node ready event."""
         # the event contains a full dump of the node
         self.data.update(event.data["nodeState"])
+        # update device config
+        if new_device_config := self.data.get("deviceConfig"):
```
kpine, line 508: Not sure this is necessary, could just re-assign it regardless, like in the constructor.

MartinHjelmare (approved):

raman325 (approved):

PR #252, opened 2021-07-13
Add new endpoint commands

raman325:
Adds `endpoint.invoke_cc_api` and `endpoint.supports_cc_api` (Upstream PR: https://github.com/zwave-js/zwave-js-server/pull/282)

As part of this change I had to move `NodeStatus` to `const` so that it could be referenced in `zwave_js_server.model.endpoint`

### DIFF
--- a/zwave_js_server/model/endpoint.py
+++ b/zwave_js_server/model/endpoint.py
@@ -113,8 +113,7 @@ class Endpoint(EventBase):
         wait_for_result: bool = None,
     ) -> Any:
         """Call endpoint.invoke_cc_api command."""
-        result = await Endpoint._async_send_command(
-            self,
+        result = await self._async_send_command(
             "invoke_cc_api",
             commandClass=command_class.value,
             methodName=method_name,
@@ -128,8 +127,7 @@ class Endpoint(EventBase):
 
     async def async_supports_cc_api(self, command_class: CommandClass) -> bool:
         """Call endpoint.supports_cc_api command."""
-        result = await Endpoint._async_send_command(
-            self,
+        result = await self._async_send_command(
             "supports_cc_api",
             commandClass=command_class.value,
             require_schema=7,
--- a/zwave_js_server/model/endpoint.py
+++ b/zwave_js_server/model/endpoint.py
@@ -82,7 +82,9 @@ class Endpoint(EventBase):
         or not based on the node status.
         """
         if self.client.driver is None:
-            raise FailedCommand("Command failed", "failed_command", "The client is not connected")
+            raise FailedCommand(
+                "Command failed", "failed_command", "The client is not connected"
+            )
         node = self.client.driver.controller.nodes[self.node_id]
         kwargs = {}
         message = {
--- a/zwave_js_server/model/endpoint.py
+++ b/zwave_js_server/model/endpoint.py
@@ -80,6 +80,7 @@ class Endpoint(EventBase):
         If wait_for_result is not None, it will take precedence, otherwise we will decide to wait
         or not based on the node status.
         """
+        assert self.client.driver
         node = self.client.driver.controller.nodes[self.node_id]
         kwargs = {}
         message = {
@@ -130,4 +131,6 @@ class Endpoint(EventBase):
             require_schema=7,
             wait_for_result=True,
         )
+        # We can assert this because we are forcing the client to await the response
+        assert result
         return cast(bool, result["supported"])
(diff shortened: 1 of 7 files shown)


raman325:
We should never hit the two lines that are uncovered. I can force a test that adds coverage, but I'm not sure that that's useful

zwave_js_server/model/endpoint.py
```diff
@@ -64,3 +67,75 @@ def installer_icon(self) -> Optional[int]:
...
+
+    async def async_invoke_cc_api(
+        self,
+        command_class: CommandClass,
+        method_name: str,
+        *args: Any,
+        wait_for_result: bool = None,
+    ) -> Any:
+        """Call endpoint.invoke_cc_api command."""
+        result = await Endpoint._async_send_command(
```
MartinHjelmare, line 116:
```suggestion
        result = await self._async_send_command(
```

zwave_js_server/model/endpoint.py
```diff
@@ -64,3 +67,75 @@ def installer_icon(self) -> Optional[int]:
...
+            require_schema=7,
+            wait_for_result=wait_for_result,
+        )
+        if result is None:
+            return None
+        return result["response"]
+
+    async def async_supports_cc_api(self, command_class: CommandClass) -> bool:
+        """Call endpoint.supports_cc_api command."""
+        result = await Endpoint._async_send_command(
```
MartinHjelmare, line 131:
```suggestion
        result = await self._async_send_command(
```

zwave_js_server/model/endpoint.py
```diff
@@ -64,3 +67,75 @@ def installer_icon(self) -> Optional[int]:
...
+        **cmd_kwargs: Any,
+    ) -> Optional[Dict[str, Any]]:
+        """
+        Send an endpoint command. For internal use only.
+
+        If wait_for_result is not None, it will take precedence, otherwise we will decide to wait
+        or not based on the node status.
+        """
+        if self.client.driver is None:
+            raise FailedCommand(
```
MartinHjelmare, line 85: The library user may cause this error to be raised so we should test it.

zwave_js_server/model/endpoint.py
```diff
@@ -64,3 +67,75 @@ def installer_icon(self) -> Optional[int]:
...
+        """Call endpoint.supports_cc_api command."""
+        result = await Endpoint._async_send_command(
+            self,
+            "supports_cc_api",
+            commandClass=command_class.value,
+            require_schema=7,
+            wait_for_result=True,
+        )
+        if result is None:
+            # We should never reach this code
```
MartinHjelmare, line 139: If this can't happen we can assert the type of the result.

MartinHjelmare (changes requested):

zwave_js_server/model/endpoint.py
raman325: The reason I forced this to be `Endpoint` is because if, for whatever reason, the `Node` picks up `_async_send_command`, the command will fail (it always has to be `endpoint.invoke_cc_api` and `endpoint.supports_cc_api`). Alternatively, we can update zwave-js-server to accept `node.invoke_cc_api`.

zwave_js_server/model/endpoint.py
raman325: I'm trying to think how the user could hit this. The nodes and endpoints won't be available until the driver is created, and I don't see a place where we set the driver back to None when disconnecting. Should we set the driver to None when disconnecting since we would recreate it when we reconnect anyway (assuming I understand the logic correctly)?

MartinHjelmare: Yes, I think we should reset the driver on disconnect and also the result futures and shutdown complete event client attributes.

zwave_js_server/model/endpoint.py
MartinHjelmare: I don't find the code that is limiting this?

raman325: it doesn't right now, but actually my tests will catch this in case it does change in the future, so we will go with the suggested approach

zwave_js_server/model/endpoint.py
raman325: oh boy. This is not as simple of a change as I thought. This might take me a bit to figure out

raman325: OK I'm not quite sure how to test the new behavior of resetting everything explicitly because it doesn't appear that the mocks are set up to hit that part of the code at the moment. I've handled the scenario in the test by clearing the driver in the test itself, but I may need some help getting the additional coverage for the client

MartinHjelmare (approved):
Looks good!

PR #253, opened 2021-07-15
Store state of firmware update progress internally

raman325:
Like we do for log config updates, storing the firmware update progress internally will make for a better user experience when we add support for firmware updates in the HA UI because the user can presumably leave the firmware update screen and come back and instantly see the latest progress update.

CC @cgarwood


raman325:
@cgarwood I'm assuming you'd ideally like to see something similar for heal network progress from our discussion on Discord a while ago?

cgarwood:
Yes, that would be awesome

MartinHjelmare (approved):

PR #260, opened 2021-07-20
Add guards for Node.async_set_value

raman325:
We can raise more informative exceptions when the value_id provided doesn't exist or the value doesn't have the appropriate options

### DIFF
--- a/zwave_js_server/model/node.py
+++ b/zwave_js_server/model/node.py
@@ -425,7 +425,7 @@ class Node(Endpoint):
             option = next(
                 (
                     option
-                    for option in options.keys()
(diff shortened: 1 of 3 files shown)


zwave_js_server/model/node.py
```diff
@@ -415,6 +422,18 @@ async def async_set_value(
             "value": new_value,
         }
         if options:
+            option = next(
+                (
+                    option
+                    for option in options.keys()
```
MartinHjelmare, line 428:
```suggestion
                    for option in options
```

MartinHjelmare (approved):

PR #265, opened 2021-07-28
Add additional constants

raman325:
As a result of this comment: https://github.com/home-assistant/core/pull/53390#discussion_r678155959

I did a sweep and found all property strings that are being referenced directly in the HA code. We should be able to remove all instances of that once this gets released


MartinHjelmare (approved):
Thanks!

PR #267, opened 2021-07-30
Add enums for sensor and meter types to match device class

raman325:
We can use the `ZwaveValue.metadata.cc_specific` map to map Multilevel Sensor and Meter CC values to a type/scale, and use that to derive the right device class/EntityDescription. I think this is a better method than using units.

Let me know if the following sensor types to device class mappings make sense. You can use https://github.com/zwave-js/node-zwave-js/blob/master/packages/config/config/sensorTypes.json as the reference to look at units.
- `MultilevelSensorType.BASIS_METABOLIC_RATE` as an energy device class
- `MultilevelSensorType.TIME` as a timestamp device class
- `MultilevelSensorType.WATER_OXIDATION_REDUCTION_POTENTIAL` as a voltage device class

### DIFF
--- a/zwave_js_server/const.py
+++ b/zwave_js_server/const.py
@@ -1,6 +1,6 @@
 """Constants for the Z-Wave JS python library."""
 from enum import Enum, IntEnum
-from typing import Dict, List
+from typing import Dict, List, Union
 
 # minimal server schema version we can handle
 MIN_SERVER_SCHEMA_VERSION = 7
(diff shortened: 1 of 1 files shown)


raman325:
I created this PR upstream to demonstrate how we could use these constants: https://github.com/home-assistant/core/pull/53744

zwave_js_server/const.py
```diff
@@ -1,6 +1,6 @@
 """Constants for the Z-Wave JS python library."""
 from enum import Enum, IntEnum
-from typing import Dict, List
+from typing import Dict, List, Union
```
firstof9, line 3: `zwave_js_server/const.py:3:0: W0611: Unused Union imported from typing (unused-import)`

MartinHjelmare (approved):
Looks good!

PR #270, opened 2021-08-08
Add CC specific utility functions

raman325:
Adds utility functions to get `ccSpecific` enums

### DIFF
--- a/zwave_js_server/const.py
+++ b/zwave_js_server/const.py
@@ -1,6 +1,6 @@
 """Constants for the Z-Wave JS python library."""
 from enum import Enum, IntEnum
-from typing import Dict, List
+from typing import Dict, List, Union
 
 # minimal server schema version we can handle
 MIN_SERVER_SCHEMA_VERSION = 8
@@ -651,6 +651,8 @@ class HeatingScale(IntEnum):
 
 
 CoolingScale = HeatingScale
+
+MeterScaleType = Union[CoolingScale, ElectricScale, GasScale, HeatingScale, WaterScale]
 
 METER_TYPE_TO_SCALE_ENUM_MAP = {
     MeterType.ELECTRIC: ElectricScale,
--- a/zwave_js_server/const.py
+++ b/zwave_js_server/const.py
@@ -1,6 +1,6 @@
 """Constants for the Z-Wave JS python library."""
 from enum import Enum, IntEnum
-from typing import Dict, List, Union
+from typing import Dict, List, Set, Type, Union
 
 # minimal server schema version we can handle
 MIN_SERVER_SCHEMA_VERSION = 8
@@ -654,7 +654,7 @@ CoolingScale = HeatingScale
 
 MeterScaleType = Union[CoolingScale, ElectricScale, GasScale, HeatingScale, WaterScale]
 
-METER_TYPE_TO_SCALE_ENUM_MAP = {
+METER_TYPE_TO_SCALE_ENUM_MAP: Dict[MeterType, Type[MeterScaleType]] = {
     MeterType.ELECTRIC: ElectricScale,
     MeterType.GAS: GasScale,
     MeterType.WATER: WaterScale,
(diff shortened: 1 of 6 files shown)


zwave_js_server/util/command_class.py
```diff
@@ -0,0 +1,55 @@
...
+def get_meter_scale_type(
+    value: Value,
+) -> Union[CoolingScale, ElectricScale, GasScale, HeatingScale, WaterScale, None]:
+    """Get the ScaleType for a given value."""
+    meter_type = get_meter_type(value)
+    if meter_type is None:
+        return None
+    scale_enum = METER_TYPE_TO_SCALE_ENUM_MAP[meter_type]
+    try:
+        return scale_enum(value.metadata.cc_specific[CC_SPECIFIC_SCALE])  # type: ignore
```
MartinHjelmare, line 42: What type error are we ignoring?

raman325: `zwave_js_server/util/command_class.py:43: error: Returning Any from function declared to return "Union[HeatingScale, ElectricScale, GasScale, HeatingScale, WaterScale, None]"`

I tried type hinting scale_enum with `IntEnum` and `EnumMeta` and that didn't help

raman325:
I can't increase test coverage without creating fake values. All the code paths that aren't tested are related to error handling. Practically speaking we should never reach them

zwave_js_server/const.py
```diff
@@ -653,6 +653,8 @@ class HeatingScale(IntEnum):
 
 CoolingScale = HeatingScale
 
+MeterScaleType = Union[CoolingScale, ElectricScale, GasScale, HeatingScale, WaterScale]
+
 METER_TYPE_TO_SCALE_ENUM_MAP = {
```
MartinHjelmare, line 658: Maybe we need to type this dict?

zwave_js_server/util/command_class.py
```diff
@@ -0,0 +1,49 @@
...
+
+def get_meter_type(value: Value) -> Union[MeterType, None]:
+    """Get the MeterType for a given value."""
+    if value.command_class != CommandClass.METER:
+        raise InvalidCommandClass(value, CommandClass.METER)
+
+    try:
+        return MeterType(value.metadata.cc_specific[CC_SPECIFIC_METER_TYPE])
+    except ValueError:
+        return None
```
MartinHjelmare, line 26: When could this happen? If the device has incorrect config?

raman325: It's probably more likely that a new type gets introduced that we don't have in the enum yet. I came across an issue when testing this in the HA PR, but maybe we should start with just letting it fail and if we need to add guards we can. Things failing will otherwise be a forcing function for us to know that something is wrong

MartinHjelmare: Do you mean we should remove the try... except here?

raman325: Oh I know what it was. It was causing a test to fail because we had an old fixture that did not have this `ccSpecific` data yet. It probably just makes more sense to fix the fixture then

zwave_js_server/const.py
raman325: this is what I get when I try:
```
zwave_js_server/const.py:659: error: Dict entry 0 has incompatible type "MeterType": "Type[ElectricScale]"; expected "MeterType": "Union[HeatingScale, ElectricScale, GasScale, WaterScale]"
zwave_js_server/const.py:660: error: Dict entry 1 has incompatible type "MeterType": "Type[GasScale]"; expected "MeterType": "Union[HeatingScale, ElectricScale, GasScale, WaterScale]"
zwave_js_server/const.py:661: error: Dict entry 2 has incompatible type "MeterType": "Type[WaterScale]"; expected "MeterType": "Union[HeatingScale, ElectricScale, GasScale, WaterScale]"
zwave_js_server/const.py:662: error: Dict entry 3 has incompatible type "MeterType": "Type[HeatingScale]"; expected "MeterType": "Union[HeatingScale, ElectricScale, GasScale, WaterScale]"
zwave_js_server/const.py:663: error: Dict entry 4 has incompatible type "MeterType": "Type[HeatingScale]"; expected "MeterType": "Union[HeatingScale, ElectricScale, GasScale, WaterScale]"
```

any suggestions on how to fix that?

MartinHjelmare: Maybe we need to use a TypedDict?

zwave_js_server/util/command_class.py
raman325: Looked into it again, and it was because of a bad interview. This was the value that was throwing the error:
```
{
            "endpoint": 3,
            "commandClass": 49,
            "commandClassName": "Multilevel Sensor",
            "property": "UNKNOWN (0x00)",
            "propertyName": "UNKNOWN (0x00)",
            "ccVersion": 0,
            "metadata": {
                "type": "number",
                "readable": true,
                "writeable": false,
                "label": "UNKNOWN (0x00)",
                "ccSpecific": {
                    "sensorType": 0,
                    "scale": 0
                }
            },
            "value": 0
        },
```
Do you still think we should avoid the guard? I can fix the `ccSpecific` values here but it won't prevent the exception from raising for users that have this issue

MartinHjelmare: If we are sure that the reason is a bad interview or similar I think we should raise a specific error letting the user know that something is wrong with the config of the device.

raman325: OK, we will have to catch that in discovery then so we can continue with discovery of other values

raman325:
I was able to test one of the failure scenarios since we have the fixture data, but can't get any further than that.

zwave_js_server/const.py
raman325: It won't recognize the enum value as a key

raman325:
@MartinHjelmare what more do we need to do on this PR to get it closed? I'd like to get this merged so I can release a new version and prepare core for the update, and then also spend more time working on the entity description PR since it's a better implementation vs our current approach

zwave_js_server/const.py
raman325: I needed to add `Type` as a wrapper 🤦🏾

zwave_js_server/util/command_class.py
raman325: With the type hint fix below I was able to remove this

raman325:
Update: decided it was OK to create fake values with known bad `meterType` and `scale` values. Even if it's not a real life scenario, it validates that we will raise an exception if such an issue occurs

MartinHjelmare (approved):
Great!

PR #272, opened 2021-08-16
Support schema 8 changes to zwave-js-server

raman325:
See https://github.com/zwave-js/zwave-js-server/pull/342 for changes

- `controller.begin_inclusion` and `controller.replace_failed_node` have been updated to match upstream changes (we also added a separate method for the default inclusion strategy because it has different inputs. We will do the same for SmartStart once it is supported)
- Two new controller commands have been added to support back and forth exchange for S2 bootstrapping
- Two new controller events that the client can use to communicate back and forth with the driver
- Two new node commands related to security
- A couple of enums and data classes

### DIFF
--- a/test/model/test_controller.py
+++ b/test/model/test_controller.py
@@ -1,6 +1,7 @@
 """Test the controller model."""
 import json
 
+from zwave_js_server.const import InclusionStrategy, SecurityClass
 from zwave_js_server.event import Event
 from zwave_js_server.model import association as association_pkg
 from zwave_js_server.model import controller as controller_pkg
@@ -128,12 +129,12 @@ async def test_begin_inclusion(controller, uuid4, mock_command):
         {"command": "controller.begin_inclusion"},
         {"success": True},
     )
-    assert await controller.async_begin_inclusion()
+    assert await controller.async_begin_inclusion(InclusionStrategy.DEFAULT)
 
     assert len(ack_commands) == 1
     assert ack_commands[0] == {
         "command": "controller.begin_inclusion",
-        "includeNonSecure": None,
+        "options": {"inclusionStrategy": InclusionStrategy.DEFAULT.value},
         "messageId": uuid4,
     }
 
@@ -214,15 +215,16 @@ async def test_replace_failed_node(controller, uuid4, mock_command):
     )
 
     node_id = 52
-    include_non_secure = True
-    assert await controller.async_replace_failed_node(node_id, include_non_secure)
+    assert await controller.async_replace_failed_node(
+        node_id, InclusionStrategy.DEFAULT
+    )
 
     assert len(ack_commands) == 1
     assert ack_commands[0] == {
         "command": "controller.replace_failed_node",
         "messageId": uuid4,
         "nodeId": node_id,
-        "includeNonSecure": include_non_secure,
+        "options": {"inclusionStrategy": InclusionStrategy.DEFAULT.value},
     }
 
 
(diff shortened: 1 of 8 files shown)


zwave_js_server/model/controller.py
```diff
@@ -9,9 +11,41 @@
...
+    client_side_auth: bool
+
+    def to_dict(self) -> InclusionGrantDataType:
+        """Return InclusionGrantDataType dict from self."""
+        return {
+            "securityClasses": self.security_classes,
+            "clientSideAuth": self.client_side_auth,
+        }
+
+    @staticmethod
```
MartinHjelmare, line 36: Make this a classmethod.
```suggestion
    @classmethod
```

zwave_js_server/model/controller.py
```diff
@@ -9,9 +11,41 @@
...
+    def from_dict(data: InclusionGrantDataType) -> "InclusionGrant":
```
MartinHjelmare, line 37:
```suggestion
    def from_dict(cls, data: InclusionGrantDataType) -> "InclusionGrant":
```

zwave_js_server/model/controller.py
```diff
@@ -9,9 +11,41 @@
...
+        """Return InclusionGrant from InclusionGrantDataType dict."""
+        return InclusionGrant(
```
MartinHjelmare, line 39:
```suggestion
        return cls(
```

zwave_js_server/model/node.py
```diff
@@ -511,6 +512,30 @@ async def async_ping(self) -> bool:
...
+        data = (
+            await self.async_send_command(
+                "has_security_class",
+                securityClass=security_class,
+                require_schema=8,
+                wait_for_result=True,
+            )
+            or {}
+        )
+        return cast(bool, data["hasSecurityClass"])
```
MartinHjelmare, line 526: There will be a KeyError if the default empty dict is returned above.

test/model/test_controller.py
```diff
@@ -572,3 +593,50 @@ async def test_statistics_updated(controller):
...
+    await controller.async_grant_security_classes(inclusion_grant)
+
+    assert len(ack_commands) == 1
+    assert ack_commands[0] == {
+        "command": "controller.grant_security_classes",
+        "messageId": uuid4,
+        "inclusionGrant": inclusion_grant_dict,
+    }
+
+    """Test grant security classes event."""
```
MartinHjelmare, line 617: Use `# ` for comments.

MartinHjelmare (commented):
Should we bump minimum server schema version to 8?

zwave_js_server/model/node.py
raman325: The reason for the empty dict here is to get past mypy. Unless the command fails, we will always get `hasSecurityClass`

raman325:
> Should we bump minimum server schema version to 8?

Yes I did this as part of this PR

zwave_js_server/model/node.py
MartinHjelmare: If we know that `async_send_command` will return a non empty dict we should just `assert data`.

zwave_js_server/model/node.py
```diff
@@ -511,6 +512,28 @@ async def async_ping(self) -> bool:
...
+        assert data
+        return cast(bool, data["hasSecurityClass"])
+
+    async def async_get_highest_security_class(self) -> SecurityClass:
+        """Get the highest security class that a node supports."""
+        data = await self.async_send_command(
+            "get_highest_security_class", require_schema=8, wait_for_result=True
+        )
+        # We should never get here
+        if not data:
```
MartinHjelmare, line 532: Should we `assert data` here too?

raman325: thanks missed that one

MartinHjelmare (approved):
Looks good!

PR #273, opened 2021-08-18
Remove dupe CC enum

raman325:
both ALARM and NOTIFICATION are assigned CC ID 113. zwave-js considers 113 to be for the notification command class so we should too


MartinHjelmare:
We use the alarm constant in a test.

MartinHjelmare (approved):
Thanks!

PR #282, opened 2021-08-27
Add Multilevel Sensor CC scale enums and util functions

raman325:
I generated new enums for all of the Multilevel Sensor types and scale maps (I used this script to generate the class definitions from the config: https://gist.github.com/raman325/b72a01bdf5cbc0bc497aab3e50726318). I used these new enums to create a new util function to get the scale type for a given value (this will let us eventually improve our unit mapping in HA).

I also restructured the util.command_class module so that the cc specific functions are split into their own submodules.

Follows #281


MartinHjelmare (approved):

MartinHjelmare:
Should we include the script in a scripts folder in this project so that we can regenerate the class definitions later?

raman325:
> Should we include the script in a scripts folder in this project so that we can regenerate the class definitions later?

Sounds good, feel free to add comments to that script too once I add it

MartinHjelmare (approved):

PR #284, opened 2021-08-27
Bump lib to 0.29.1

raman325:
We should bump the lib in core to ~~fix the incorrect classification of `ElectricScale.PULSE_COUNT` that's fixed in #283 (we can save adding support for the `gas` device class for later)~~ improve the `UnknownValueData` exception message #284 to reduce the number of tickets that could get opened as a result of seeing this message in the logs

Because of the refactor to constants, this dependency update will be a bit of work on the core side but it shouldn't be too bad


raman325:
@MartinHjelmare so are we good to merge here?

MartinHjelmare (approved):
Yes! :+1:

PR #285, opened 2021-08-30
Improve UnknownValueData exception messaging

raman325:
During the HA beta, it became apparent that our current messaging was a little too strongly worded and made users think there was a bug. I have softened the language a bit and given the user a potential action they can take to resolve the issue.

I'm worried that when 2021.9 gets released, we will get a ton of GH issues opened related to this exception. Thoughts on merging https://github.com/home-assistant-libs/zwave-js-server-python/pull/284 after merging this PR and doing a dependency bump in HA ahead of that so that we can divert some of that? I originally thought we should do the bump to correct the misclassification of the `PULSE_COUNT` enum, but we can handle that in HA code so a bump is not strictly needed for that. It is for this though

### DIFF
--- a/zwave_js_server/exceptions.py
+++ b/zwave_js_server/exceptions.py
@@ -151,6 +151,8 @@ class UnknownValueData(BaseZwaveJSServerError):
         self.path = path
         super().__init__(
             f"Value {value} has unknown data in the following location: {path}. "
-            "Please report this issue, it can be caused by either an upstream issue "
-            "with the driver or missing support for this data in the library"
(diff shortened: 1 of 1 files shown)


MartinHjelmare (approved):

MartinHjelmare:
Is the move of the constants a breaking change so we need to update the integration if we bump now?

raman325:
> Is the move of the constants a breaking change so we need to update the integration if we bump now?

Yes we'd have to fix imports but there are no code changes required (other than the rename of the enum from PULSE to PULSE_COUNT

MartinHjelmare:
Ok, sounds good.

PR #297, opened 2021-09-21
Wait for result when fetching lock usercodes

raman325:
Noticed this the other day but forgot to submit a PR. We should always wait for a result from the lock in this util function because we expect the data to be there as part of the return


MartinHjelmare (approved):
Thanks!

PR #298, opened 2021-09-21
Bump min server version and lib version

raman325:
Pending an upstream release


setup.py
```diff
@@ -5,7 +5,7 @@
 
 PROJECT_DIR = Path(__file__).parent.resolve()
 README_FILE = PROJECT_DIR / "README.md"
-VERSION = "0.30.0"
+VERSION = "0.30.1"
```
MartinHjelmare, line 8: Should we make it a minor version bump since we're bumping the minimum server schema version?

MartinHjelmare (approved):

PR #302, opened 2021-10-02
Add highest security class property to node

raman325:
Added the property to the state dump here: https://github.com/zwave-js/zwave-js-server/pull/402

### DIFF
--- a/zwave_js_server/model/node.py
+++ b/zwave_js_server/model/node.py
@@ -134,6 +134,7 @@ class NodeDataType(EndpointDataType):
     commandClasses: List[CommandClassInfoDataType]
     values: List[ValueDataType]
     statistics: NodeStatisticsDataType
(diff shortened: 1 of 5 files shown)


zwave_js_server/model/node.py
```diff
@@ -134,6 +134,7 @@ class NodeDataType(EndpointDataType):
     commandClasses: List[CommandClassInfoDataType]
     values: List[ValueDataType]
     statistics: NodeStatisticsDataType
+    highestSecurityClass: SecurityClass
```
MartinHjelmare, line 137: Is the type correct? Isn't this just the integer value?

raman325: yes, fixing this here and in the other two places I made this mistake

MartinHjelmare (approved):
Thanks!

PR #303, opened 2021-10-02
Increase max message size for dumps to support large networks

raman325:
See https://discord.com/channels/330944238910963714/800356888827002880/893653844813299713 and https://github.com/zwave-js/zwavejs2mqtt/issues/1769 for context. We should push a new release with this change as upstream changes in a recent version of zwave-js caused the value list to increase significantly for some devices

Will there be any issues on the HA side for large dumps? I took a look at the `HomeAssistantView` logic and I don't see any limits being applied.

### DIFF
--- a/zwave_js_server/client.py
+++ b/zwave_js_server/client.py
@@ -8,7 +8,11 @@ from typing import Any, Dict, Optional, cast
 
 from aiohttp import ClientSession, ClientWebSocketResponse, WSMsgType, client_exceptions
 
-from .const import MAX_SERVER_SCHEMA_VERSION, MIN_SERVER_SCHEMA_VERSION
+from .const import (
+    MAX_NETWORK_STATE_SIZE,
+    MAX_SERVER_SCHEMA_VERSION,
+    MIN_SERVER_SCHEMA_VERSION,
+)
 from .event import Event
 from .exceptions import (
     CannotConnect,
@@ -102,6 +106,7 @@ class Client:
             self._client = await self.aiohttp_session.ws_connect(
                 self.ws_server_url,
                 heartbeat=55,
+                max_msg_size=MAX_NETWORK_STATE_SIZE,
             )
         except (
             client_exceptions.WSServerHandshakeError,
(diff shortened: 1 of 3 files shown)


kpine:
You're missing the client connection.

https://github.com/home-assistant-libs/zwave-js-server-python/blob/63ccaba7fe13d7d16bfda25b3755cea802af67dc/zwave_js_server/client.py#L102-L105

HA issue is https://github.com/home-assistant/core/issues/56922.

raman325:
I didnt think it was needed but now I am realizing this would happen during connect because of the start listening command. I'll fix it in the morning

MartinHjelmare:
@balloob do you see any issues with larger the ws message size limit?

zwave_js_server/client.py
```diff
@@ -102,6 +106,7 @@ async def connect(self) -> None:
             self._client = await self.aiohttp_session.ws_connect(
                 self.ws_server_url,
                 heartbeat=55,
+                max_msg_size=MAX_NETWORK_STATE_SIZE,
```
balloob, line 109: Why add it here? Don't we only need it in `dump.py` ?

kpine: No, it's on any client connection.

balloob:
Increasing size should be fine. We trust the other side, which is a ZJS server. 

This is however not a fix, but a workaround. Should we look into doing breaking dumps in multiple pieces? (Although, ofc, then it's not a perfect snapshot)

firstof9:
> Should we look into doing breaking dumps in multiple pieces?

Is compression an option at all?

zwave_js_server/client.py
raman325: see https://github.com/home-assistant-libs/zwave-js-server-python/pull/303#issuecomment-932694444

raman325:
We either need to break the dump into multiple pieces or use compression. Neither is a trivial change, and this is a faster interim solution.

raman325:
With @balloob and @kpine 's help, this PR will add compression support to the server and will compress the `start_listening` command response as long as the client supports it,. This is a much simpler change to add to the lib so closing this one

PR #314, opened 2021-11-03
Add context object to log message model

raman325:
Implements the feature added here: https://github.com/zwave-js/node-zwave-js/pull/3251

This feature would allow clients of the library to selectively filter log messages. Imagine being able to pick an entity in HA and just watch the logs for the value associated with it.


MartinHjelmare:
Is there any documentation in the zwave-js docs about the log messages interface?

raman325:
I don't believe so. Perhaps something to talk to Al about

MartinHjelmare (approved):
Looks good!

PR #315, opened 2021-11-03
Handle upstream changes and new commands for Smart Start

raman325:
Related to https://github.com/zwave-js/zwave-js-server/pull/432

- Adds support for the new SmartStart provisioning workflows as well as using the S2 inclusion workflow using a QR code. 
- Adds a command to be able to call `parseQRCodeString`.
- Update `beginExclusion` command to include new option.

One thing that I have done is moved the S2 security `beginInclusion` command into its own method given the multiple ways that that workflow can work. The existing `async_begin_inclusion` method will still work for S2 security as the inclusion strategy but does not support all of the various options that are now available in the linked PR

### DIFF
--- a/zwave_js_server/model/controller.py
+++ b/zwave_js_server/model/controller.py
@@ -539,20 +539,52 @@ class Controller(EventBase):
     async def async_replace_failed_node(
         self,
         node_id: int,
-        inclusion_strategy: Union[
-            Literal[InclusionStrategy.SECURITY_S0],
-            Literal[InclusionStrategy.SECURITY_S2],
-            Literal[InclusionStrategy.INSECURE],
+        inclusion_strategy: Literal[
+            InclusionStrategy.DEFAULT,
+            InclusionStrategy.SECURITY_S0,
+            InclusionStrategy.SECURITY_S2,
+            InclusionStrategy.INSECURE,
         ],
+        force_security: Optional[bool] = None,
+        provisioning: Optional[
+            Union[str, ProvisioningEntry, QRProvisioningInformation]
+        ] = None,
     ) -> bool:
         """Send replaceFailedNode command to Controller."""
+        # Most functionality was introduced in Schema 8
+        require_schema = 8
+        options: Dict[str, Any] = {"strategy": inclusion_strategy}
+        # forceSecurity can only be used with the default inclusion strategy
+        if force_security is not None:
+            if inclusion_strategy != InclusionStrategy.DEFAULT:
+                raise ValueError(
+                    "`forceSecurity` option is only supported with inclusion_strategy=DEFAULT"
+                )
+            options["forceSecurity"] = force_security
+
+        # provisioning can only be used with the S2 inclusion strategy and may need
+        # additional processing
+        if provisioning is not None:
+            if inclusion_strategy != InclusionStrategy.SECURITY_S2:
+                raise ValueError(
+                    "`provisioning` option is only supported with inclusion_strategy=SECURITY_S2"
+                )
+            # Provisioning option was introduced in Schema 11
+            require_schema = 11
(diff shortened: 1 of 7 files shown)


zwave_js_server/model/controller.py
```diff
@@ -42,6 +52,117 @@ def from_dict(cls, data: InclusionGrantDataType) -> "InclusionGrant":
...
+            ]
+        return data
+
+    @classmethod
+    def from_dict(
+        cls, data: QRProvisioningInformationDataType
+    ) -> "QRProvisioningInformation":
+        """Return QRProvisioningInformation from QRProvisioningInformationDataType dict."""
+        return cls(
+            QRCodeVersion(data["version"]),
```
MartinHjelmare, line 147: Maybe use keyword arguments for clarity?

zwave_js_server/model/controller.py
```diff
@@ -295,6 +416,31 @@ async def async_begin_inclusion(
         )
         return cast(bool, data["success"])
 
+    async def async_begin_inclusion_s2(
```
MartinHjelmare, line 419: Can't we merge this with the existing method and raise a RuntimeError or similar if it's called with the wrong combination of arguments, ie provisioning_info and not S2 strategy?

MartinHjelmare: Yeah. I think we should try to stay in line with the zwave-js api as far as possible.

raman325: shall we remove the `async_begin_inclusion_default` method? For now I've added a deprecation warning. Maybe we leave it there for a bit and remove it in the future?

zwave_js_server/model/controller.py
```diff
@@ -401,43 +412,45 @@ def heal_network_progress(self) -> Optional[Dict[int, str]]:
...
+                provisioning = provisioning.to_dict()
+            options["provisioning"] = provisioning
+
         data = await self.client.async_send_command(
             {
                 "command": "controller.begin_inclusion",
                 "options": options,
             },
-            require_schema=11,
+            require_schema=8,
```
MartinHjelmare, line 453: Shouldn't we set the minimum version depending on what options we use?

zwave_js_server/model/controller.py
MartinHjelmare: I think we can make a breaking change. We're bumping the minimum server version so that's a breaking change anyway.

zwave_js_server/model/controller.py
```diff
@@ -279,38 +408,107 @@ def heal_network_progress(self) -> Optional[Dict[int, str]]:
...
+        await self.client.async_send_command(
+            {
+                "command": "controller.provision_smart_start_node",
+                "entry": provisioning_info
+                if isinstance(provisioning_info, str)
+                else provisioning_info.to_dict(),
+            },
+            require_schema=11,
+        )
+        return None
```
MartinHjelmare, line 474: It doesn't look like we need to return a value explicitly here.
```suggestion
```

zwave_js_server/model/controller.py
```diff
@@ -279,38 +408,107 @@ def heal_network_progress(self) -> Optional[Dict[int, str]]:
...
+    ) -> None:
+        """Send unprovisionSmartStartNode command to Controller."""
+        await self.client.async_send_command(
+            {
+                "command": "controller.unprovision_smart_start_node",
+                "dskOrNodeId": dsk_or_node_id,
+            },
+            require_schema=11,
+        )
+        return None
```
MartinHjelmare, line 488:
```suggestion
```

MartinHjelmare (dismissed):
Looks good!

MartinHjelmare (approved):

PR #316, opened 2021-11-12
Remove time multisensor type from timestamp set

raman325:
As discussed, removing the type from the set will prevent the device class from being improperly set up stream. Since it is empty we can remove it entirely.


MartinHjelmare (approved):
Thanks!

Note that this is a breaking change. We'll need to remove the use here:
https://github.com/home-assistant/core/blob/21f92f628698d1e42dd713cad676081b97cf0e5f/homeassistant/components/zwave_js/discovery_data_template.py#L80

raman325:
right - I was originally going to leave the variable in place as an empty set but then thought that wouldn't be very useful since the spec doesn't change often so who knows when we'd actually have something to put there

PR #317, opened 2021-11-12
Improve multilevel sensor constants script and add units map

raman325:
Adding mappings for units across sensor types so that we can map them to units in HA. I updated the generation script to accomplish this. While I was there, I updated it to clean/speed it up. I also rewrote it so that there's no user input required to update the actual constants module (it is done automatically and with idempotency, even for the code that was manually added)

EDIT: This looks like a bigger change than it is because building idempotency required me to make things alphabetical so we don't generate random commits because of upstream reordering. The relevant additions are here (there are otherwise no functional changes): https://github.com/home-assistant-libs/zwave-js-server-python/pull/317/files#diff-8e38760969b747004856e9b341d87a4cad2e6d4535ad91cc9fbb27bd3ec0936eR684


zwave_js_server/const/command_class/multilevel_sensor.py
```diff
@@ -370,155 +376,155 @@ class MoistureScale(IntEnum):
...
-    """Enum with all known scales for WATER_CHLORINE_LEVEL multilevel sensor type."""
+    """Enum with all known scales for WATER_CHLORINE_LEVEL multilevel sensor types."""
 
     # https://github.com/zwave-js/node-zwave-js/blob/master/packages/config/config/sensorTypes.json
     MILLIGRAM_PER_LITER = 0
 
 
 class WaterOxidationReductionPotentialScale(IntEnum):
-    """Enum with all known scales for WATER_OXIDATION_REDUCTION_POTENTIAL multilevel sensor type."""
+    """Enum with all known scales for WATER_OXIDATION_REDUCTION_POTENTIAL multilevel sensor types."""
```
firstof9, line 520: Linter says: `C0301: Line too long (101/100) (line-too-long)`

scripts/generate_multilevel_sensor_constants.py
```diff
@@ -40,46 +50,53 @@ def normalize_name(name: str) -> str:
...
 
 sensors = {}
 for sensor_id, sensor_props in sensor_types.items():
     sensor_id = int(sensor_id, 16)
     scale_def = sensor_props["scales"]
-    remove_paranthesis_ = True
+    remove_paranthesis_: bool = True
     if sensor_id in (87, 88):
-        remove_paranthesis_ = False
+        remove_paranthesis_: bool = False
```
MartinHjelmare, line 99: Why is the type needed? Normally those are determined automatically.

scripts/generate_multilevel_sensor_constants.py
```diff
@@ -90,69 +107,142 @@ def normalize_scale_definition(scale_definitions: dict[str, dict]) -> dict[str,
...
+    ]
+)
 
-scale_class_names = [
-    f"{format_for_class_name(scale_name)}Scale" for scale_name in scales
-]
-print(f"MultilevelSensorScaleType = Union[{', '.join(sorted(scale_class_names))}]")
-print()
-print()
+with open(MULTILEVEL_SENSOR_CONST_FILE_PATH, "r", encoding="utf-8") as fp:
```
MartinHjelmare, line 220: This can probably make it one line: https://docs.python.org/3/library/pathlib.html#pathlib.Path.read_text

scripts/generate_multilevel_sensor_constants.py
```diff
@@ -13,12 +16,19 @@
 SENSOR_TYPES_FILE_PATH = "packages/config/config/sensorTypes.json"
 DEFAULT_SCALES_FILE_PATH = "packages/config/config/scales.json"
 
+base_path = os.path.dirname(__file__)
+
+
+MULTILEVEL_SENSOR_CONST_FILE_PATH = os.path.join(
```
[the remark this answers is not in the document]
raman325: thanks, never used it before but it offers some convenient ways to do these operations

scripts/generate_multilevel_sensor_constants.py
raman325: my editor was complaining but when I ran mypy it didn't so I have removed them

MartinHjelmare (approved):
Nice!

scripts/generate_multilevel_sensor_constants.py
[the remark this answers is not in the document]
MartinHjelmare: Yeah, it's perfect when writing scripts.

PR #319, opened 2021-11-13
Add unit sets for meters

raman325:
Since we know the units for meters, we can also create an HA unit to zwave unit mapping.

I also added types to the constants I created in #317


MartinHjelmare (approved):
Looks good!

PR #323, opened 2021-11-18
Add controller.supports_feature cmd and additional validations

raman325:
We've added support for the `controller.supports_feature` command and add additional validations based on Discord conversations and the review from https://github.com/home-assistant/core/pull/59037


raman325:
we should also add all the validation we are adding in the WS API to the lib, will add that here

kpine:
Why does this need to be known?

raman325:
Generally speaking we try to handle error scenarios within the lib rather than pushing it to the driver

kpine:
But what about all the other errors that can occur? Why stop at the string length?

raman325:
If there are easy failure scenarios to handle we should. I can add the 90 prefix as a check for example. I guess it's a grey area as to where we stop but I don't think a PR that added more validation would be rejected

test/model/test_controller.py
```diff
@@ -306,6 +318,52 @@ async def test_begin_inclusion_errors(controller):
...
+    provisioning_entry = controller_pkg.QRProvisioningInformation(
+        QRCodeVersion.SMART_START,
+        [SecurityClass.S2_UNAUTHENTICATED],
+        "test",
+        1,
+        1,
+        1,
+        1,
+        1,
+        1,
```
balloob, line 335: Make these values unique (ie incremental) so you can verify the inputs are mapped to correct msg key. Using keyword args here would also help verify that things work. Having so many positional args is easy to mess up.

zwave_js_server/model/controller.py
```diff
@@ -442,10 +449,27 @@ async def async_begin_inclusion(
             require_schema = 11
             # String is assumed to be the QR code string so we can pass as is
             if isinstance(provisioning, str):
+                if (
+                    len(provisioning) < MINIMUM_QR_STRING_LENGTH
+                    or provisioning[:2] != "90"
```
balloob, line 454:
```suggestion
                    or not provisioning.startswith("90")
```

zwave_js_server/model/controller.py
```diff
@@ -442,10 +449,27 @@ async def async_begin_inclusion(
...
+                ):
+                    raise ValueError(
+                        f"QR code string must be at least {MINIMUM_QR_STRING_LENGTH} characters "
+                        "long and start with `90`"
+                    )
                 options["provisioning"] = provisioning
             # Otherwise we assume the data is ProvisioningEntry or
             # QRProvisioningInformation
             else:
```
balloob, line 463: can be an `elif` ?

zwave_js_server/model/controller.py
```diff
@@ -442,10 +449,27 @@ async def async_begin_inclusion(
...
+                if (
+                    isinstance(provisioning, QRProvisioningInformation)
+                    and provisioning.version == QRCodeVersion.SMART_START
+                ):
+                    await self.async_provision_smart_start_node(provisioning)
```
balloob, line 468: You start something and then raise ?

zwave_js_server/model/controller.py
balloob: Nope can't.

balloob (approved):

zwave_js_server/model/controller.py
raman325: Yes - what this does is stores the provisioning entry in the controller. What would you suggest? Just raise and let the caller do the right thing?

balloob: Let's just do the right thing.

raman325:
one departure from our conversation on Discord is that instead of calling `controller.provision_smart_start_node` automatically during an inclusion process when a Smart Start QR code is provided, we raise an exception. It is up to the user of the library to detect when a QR code is a Smart Start one and to call `provision_smart_start_node` instead of `begin_inclusion`

test/model/test_controller.py
```diff
@@ -373,23 +405,41 @@ async def test_provision_smart_start_node_qr_info(controller, uuid4, mock_comman
...
+        3,
+        4,
+        5,
+        6,
+        "test2",
+        7,
+        "test3",
+        [Protocols.ZWAVE],
+    )
+    with pytest.raises(ValueError):
```
MartinHjelmare, line 440: It would be good to add a comment above each block that tests a ValueError mentioning what we test for. The ValueError doesn't tell us that by itself.

test/model/test_utils.py
```diff
@@ -54,3 +58,7 @@ async def test_parse_qr_code_string(client, mock_command, uuid4):
         == 1
     )
     assert qr_provisioning_info.supported_protocols == [Protocols.ZWAVE]
+
+    # Test invalid QR code length fails
```
MartinHjelmare, line 62: This is perfect.

PR #329, opened 2021-12-03
Support additional properties for QRProvisioningInformation

raman325:
After talking to Al, I realized that additional properties like `name` and `location` can be added to both a `ProvisioningEntry` and a `QRProvisioningInformation`. In some ways, you could consider `QRPRovisioningInformation` as a subclass of `ProvisioningEntry` with some additionally defined keys, but because it's a dataclass with an optional attribute, we can't actually create subclasses so we will just duplicate the logic in both classes


MartinHjelmare:
If we want to subclass we can create a mixin for the required keys in the subclass.

MartinHjelmare (approved):
Looks good!

PR #337, opened 2021-12-23
Add new get state commands and update internal states accordingly

raman325:
In https://github.com/zwave-js/zwave-js-server/pull/474 we are introducing new commands to get the node state and controller state, as well as a couple of new node commands to update attributes. While I implemented this, I thought it also made sense to update our internal state when we retrieve a new state for the controller or node, but this has a domino effect on updates that need to happen to our various model class instances in order for instance references to stay valid. I still need to add tests but I wanted to put this up in draft so we could discuss the approach


zwave_js_server/model/endpoint.py
```diff
@@ -68,6 +67,17 @@ def user_icon(self) -> Optional[int]:
...
         return self.data.get("userIcon")
 
+    def update(
+        self,
+        data: EndpointDataType,
+        values: Optional[Dict[str, Union[ConfigurationValue, Value]]] = None,
+    ) -> None:
+        """Update the endpoint data."""
+        self.data = data
+        for value_id, value in (values or {}).items():
```
MartinHjelmare, line 77:
```suggestion
        if values is None:
            return
        for value_id, value in values.items():
```

zwave_js_server/model/endpoint.py
```diff
@@ -68,6 +67,17 @@ def user_icon(self) -> Optional[int]:
         """Return user icon property."""
         return self.data.get("userIcon")
 
+    def update(
```
MartinHjelmare, line 70: When will this be called besides on init?

zwave_js_server/model/endpoint.py
```diff
@@ -68,6 +67,17 @@ def user_icon(self) -> Optional[int]:
...
+        self,
+        data: EndpointDataType,
+        values: Optional[Dict[str, Union[ConfigurationValue, Value]]] = None,
+    ) -> None:
+        """Update the endpoint data."""
+        self.data = data
+        for value_id, value in (values or {}).items():
+            if value_id not in self.values:
```
MartinHjelmare, line 78: If the id is already there, how will the value be updated when state is updated?

zwave_js_server/model/node.py
```diff
@@ -550,6 +530,48 @@ def highest_security_class(self) -> Optional[SecurityClass]:
...
+    def is_controller_node(self) -> bool:
+        """Return whether the node is a controller node."""
+        return self.data["isControllerNode"]
+
+    @property
+    def keep_awake(self) -> bool:
+        """Return whether the node is set to keep awake."""
+        return self.data["keepAwake"]
+
+    def update_state(self, data: NodeDataType) -> None:
```
MartinHjelmare, line 543: In the other models we call it `update`.

zwave_js_server/model/controller.py
```diff
@@ -807,6 +831,73 @@ async def async_supports_feature(self, feature: ZwaveFeature) -> Optional[bool]:
         )
         return cast(Optional[bool], data.get("supported"))
 
+    async def async_get_state(self) -> ControllerDataType:
+        """Get controller state."""
+        data = await self.client.async_send_command(
+            {"command": "controller.get_state"}, require_schema=14
+        )
+        self.update(data["state"])
+        return self.data
```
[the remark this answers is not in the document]
raman325: the controller state now includes `inclusionState`. The frontend should call this when pulling up the configuration panel to determine whether a device is already in the process of inclusion or not

zwave_js_server/model/node.py
raman325: I was getting typing errors because the signature for this update method and the parent update method for endpoint were different. This was a lazy solution to fixing the problem

zwave_js_server/model/endpoint.py
raman325: the value stored on the endpoint is the same instance as the one stored on the node. The node updates the existing values so we just need to make sure our dictionary has all of the values

zwave_js_server/model/node.py
MartinHjelmare: That is a sign that our design is not optimal. Let's think about this a bit.

raman325: maybe we shot ourselves in the foot by making Node a subclass of Endpoint. It does follow the zwave-js model to do so but from a practical standpoint I'm not sure we get a lot of benefits from modeling it this way

zwave_js_server/model/controller.py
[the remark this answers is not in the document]
raman325: alternatively we don't return the data, just update state, and access the data via the model in a second call. Is there a preferred approach?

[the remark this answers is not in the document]
MartinHjelmare: That sounds better.

zwave_js_server/model/node.py
raman325: so to follow up on this, there are two options in my mind:
1. We match the signature of the update function on both classes. This would mean passing the values separately from the rest of the node state data into the update function, which is fine but a little ugly in my opinion.
2. We break the relationship between `Endpoint` and `Node` and with them being independent, there is no problem

Neither solution is optimal in my opinion, but I am sort of leaning towards 1 because it's less of a model break and more of a convention break. Thoughts?

raman325: Actually 1 would create issues because in Endpoint we pass `Value`/`ConfigurationValue` whereas in Node we get `ValueDataType`. We can't pass `Value`/`ConfigurationValue` into node because we want the node to update the value data accordingly, and we can't pass `ValueDataType` into `Endpoint` because then we'd be creating completely new `Value` instances and we should have the instances be the same on both `Endpoint` and `Node`.

OK so adding two more options (assuming we want to pursue 1):
1. We update the typing of the update method to be `List[Union[ValueDataType, Value, ConfigurationValue]]` and assert the types in each class's update method
2. We always pass ValueDataType to both `Endpoint` and `Node`. Since `Node` is always updated first, the `Endpoint` can retrieve the `Value` from the node using the `ValueDataType` data.

1 seems cleaner to me but it may not address your concerns with the model?

raman325: The typing will not work no matter what I try for option 1 (trying to keep the subclass relationship). I think the best thing to do at this point is to break the relationship and make them independent classes.

raman325:
OK so this PR works if we:
- Move some of the classes out of the `node` module into separate `modules`
- Revert the change to make `Node` a subclass of `Endpoint` and add tests to maintain coverage

Given how much complexity that adds to this PR, I will submit those as two separate PRs which we can merge and rebase this off of to keep this one simpler

MartinHjelmare (approved):
Looks good!

PR #340, opened 2022-01-05
Add script to run a mock server from a network state dump

raman325:
It started with this message: https://discord.com/channels/330944238910963714/800356888827002880/928031685570859078

I thought it would be nice to be able to run a mock server from a network state dump so that I could emulate another user's network in HA. Right now it's pretty basic but I think there's a lot more we can add, although it probably needs some refactoring to make that work as I kind of just hacked this together.

Things that could be improved:
- Is there a better way to structure this? That goes for generally and for the ideas mentioned below
- ~~I can't figure out how to send the version information on connection without hacking it together. Any advice on how to improve this?~~ Resolved, I was doing something dumb.

Ideas:
- [x] Build a mechanism to be able to dump all messages (events, commands sent, command responses, etc.) in the library. I was thinking the following format:
```json
[
  {
    "timestamp": "2018-01-01T00:00:00.000Z",
    "record_type": "event",
    "type": "blah",
    "data":{
      "event": "blah",
      ...
    }
  },
  {
    "record_type": "command",
    "command": "blah",
    "timestamp": "2018-01-01T00:00:00.000Z",
    "response_timestamp": "2018-01-01T00:00:01.000Z",
    "command_msg": {
      "command": "blah",
      ...
    },
    "response_msg": {
      "success": true,
      "result": "blah",
      ...
    }
  },
  ...
]
```
- [x] Add support to be able to replay messages:
  - Replay events
  - Mock responses to commands (so we can test HA functionality)


scripts/run_mock_server.py
```diff
@@ -0,0 +1,185 @@
...
+
+    Only designed to handle one client at a time.
+    """
+
+    def __init__(self, network_state_dump: List[Dict[str, Any]]) -> None:
+        """Initialize class."""
+        self.network_state_dump = network_state_dump
+        self.app = web.Application()
+        self.app.add_routes([web.get("/", self.websocket_handler)])
+        self.primary_ws_resp: web.WebSocketResponse = None
```
MartinHjelmare, line 34:
```suggestion
        self.primary_ws_resp: Optional[web.WebSocketResponse] = None
```

scripts/run_mock_server.py
```diff
@@ -0,0 +1,185 @@
...
+        """Send message."""
+        await self.send_json({**data, "messageId": message_id})
+
+    async def send_success_command_response(
+        self,
+        result: Optional[Dict[str, Any]],
+        message_id: str,
+    ) -> None:
+        """Send success message."""
+        if not result:
```
MartinHjelmare, line 55:
```suggestion
        if result is None:
```

scripts/run_mock_server.py
```diff
@@ -0,0 +1,185 @@
...
+                        },
+                        message_id,
+                    )
+                elif cmd == "start_listening":
+                    await self.send_json(self.network_state_dump[2])
+                else:
+                    raise ExitException(f"Unknown data received: {data}")
+            elif msg.type == WSMsgType.ERROR:
+                logging.error(
+                    "ws connection closed with exception %s",
```
MartinHjelmare, line 121:
```suggestion
                    "Connection closed with exception %s",
```

scripts/run_mock_server.py
```diff
@@ -0,0 +1,185 @@
...
+                    self.primary_ws_resp.exception(),
+                )
+
+        logging.info("websocket connection closed")
```
MartinHjelmare, line 125:
```suggestion
        logging.info("Connection closed")
```

raman325:
I'm pretty close to having event replay and command response mocking support built into the mock server but it required some significant rework so will move this back to draft until I've had a chance to finalize it

raman325:
OK so I have gone ahead and implemented event replaying and command response mocking. Someone running this script can pass in events and commands to mock - the events will be processed in order immediately after connection and the commands will be responded to in the order that the associated commands were called from the dump.

I also added two routes to replay events on demand and to add additional command responses on demand.

One quirkiness I just realized is that we pop responses off the command list when we get a matching message, with the idea being that you may want to call the same command twice and see different responses. But that means over time the command responses will drain until there are none left. On the other hand, events don't get popped off so the same events will always replay on a new connection. I originally popped them but realized that it would be annoying to have to restart the server every time you wanted to test some event replays.

I also realized while generating the event/command message dump that the schema above is a bit redundant. Will probably refactor to simplify:
```json
[
  {
    "record_type": "event",
    "type": "value updated",
    "data": {
      "type": "event",
      "event": {
        "source": "node",
        "event": "value updated",
        "nodeId": 31,
        "args": {
          "commandClassName": "Binary Switch",
          "commandClass": 37,
          "endpoint": 0,
          "property": "currentValue",
          "newValue": false,
          "prevValue": true,
          "propertyName": "currentValue"
        }
      }
    }
  },
  {
    "record_type": "command",
    "command": "node.set_value",
    "command_msg": {
      "command": "node.set_value",
      "messageId": "634ff515326c40c2aac422330777b31d",
      "nodeId": 31,
      "value": true,
      "valueId": {
        "ccVersion": 1,
        "commandClass": 37,
        "commandClassName": "Binary Switch",
        "endpoint": 0,
        "metadata": {
          "label": "Target value",
          "readable": true,
          "type": "boolean",
          "valueChangeOptions": ["transitionDuration"],
          "writeable": true
        },
        "property": "targetValue",
        "propertyName": "targetValue",
        "value": false
      }
    },
    "response_msg": {
      "success": true,
      "type": "result",
      "result": {
          "success": true
      }
    }
  }
]
```

scripts/run_mock_server.py
```diff
@@ -0,0 +1,398 @@
...
+            return
+
+        if isinstance(data, dict):
+            if "command" not in data:
+                raise ExitException(f"Malformed command response: {data}")
+            add_command_response(self.command_responses, data)
+            await self.send_json({"success": True})
+
+
+def hashable_list(lst: list) -> tuple:
```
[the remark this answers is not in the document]
raman325: agreed, I named them before I knew exactly what I was trying to accomplish

scripts/run_mock_server.py
```diff
@@ -0,0 +1,398 @@
...
+    Class to represent a mock zwave-js-server instance.
+
+    Only designed to handle one client at a time.
+    """
+
+    def __init__(
+        self,
+        network_state_dump: List[dict],
+        events_to_replay: List[dict],
+        command_responses: DefaultDict[str, DefaultDict[HashableDict, list]],
```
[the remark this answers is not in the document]
raman325: events come from zwave-js whereas commands are initiated by the client, aka the lib or HA. It only makes sense to emulate a response when it's initiated by the lib or HA. In the case of the lib, we could theoretically use this in tests, in which case a list would work because we are reproducing the same behavior over and over again. But if this is to help one of us troubleshoot an issue using the HA frontend, I didn't want to force us to have to perform the exact sequence in the order that the user did in order to conduct the test. This still preserves order but allows us flexibility in case some commands are no-op to the problem, or the order doesn't matter.

With all of that being said, I don't know why I went for a nested dictionary when a single dictionary would have worked. I will refactor accordingly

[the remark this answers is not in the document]
MartinHjelmare: Ok. Yeah, if it's not usable in the manual test it won't matter how close we are to the real case in the test. I guess we need to have a balance.

scripts/run_mock_server.py
```diff
@@ -0,0 +1,398 @@
...
+            for record in data:
+                self.send_json(record)
+            return
+        if isinstance(data, dict):
+            if "event" not in data:
+                raise ExitException(f"Malformed event: {data}")
+            self.send_json(data)
+
+    @response_decorator()
+    async def add_command_response_handler(self, data: Union[dict, list]) -> None:
```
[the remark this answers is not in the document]
raman325: I also realized that this endpoint doesn't need to be a websocket server which simplifies its logic significantly

scripts/run_mock_server.py
```diff
@@ -0,0 +1,350 @@
...
+        """Send success message."""
+        if result is None:
+            result = {}
+        await self.send_command_response(
+            {"result": result, "type": "result", "success": True}, message_id
+        )
+
+    async def process_record(self, record: dict) -> None:
+        """Process a replay dump record."""
+        if record["record_type"] == "event":
```
MartinHjelmare, line 93: Should we make a guard clause instead and require the record_type key with value event or command, otherwise raise ExitException? Otherwise there could be a KeyError here perhaps?

scripts/run_mock_server.py
```diff
@@ -0,0 +1,350 @@
...
+
+    events_to_replay = []
+    command_responses: DefaultDict[HashableDict, list] = defaultdict(list)
+
+    if args.combined_replay_dump_path:
+        with open(args.combined_replay_dump_path, "r", encoding="utf8") as fp:
+            records: List[dict] = json.load(fp)
+
+            for record in records:
+                if record["record_type"] == "event":
```
MartinHjelmare, line 293: Maybe it's enough to require the record_type key here? I guess we'll get a KeyError here already.

scripts/run_mock_server.py
raman325: I improved guards in both locations. With that being said, I think we should ignore any additional lack of guards for now and focus on functionality. We can always add in more guards later and because this is a developer tool, it's on us to make sure that both the method to generate these replay dumps and the server logic work well in tandem.

scripts/run_mock_server.py
raman325: I improved guards in both locations. With that being said, I think we should ignore any additional lack of guards for now and focus on functionality. We can always add in more guards later and because this is a developer tool, it's on us to make sure that both the method to generate these replay dumps and the server logic work well in tandem.

raman325:
TODO: handle commands we don't wait for a response for. Right now we will fail to process the dump because we expect a result

MartinHjelmare (approved):

PR #341, opened 2022-01-06
Add constants for Humidity Control CCs

EiNSTeiN-:
This PR adds the constants that are necessary to use the Humidity Control CCs which are added by https://github.com/zwave-js/node-zwave-js/pull/4048


firstof9:
I think tests will be required prior to merging. Just fyi.

EiNSTeiN-:
This PR is adding constants, how are these tested for other CCs? I don't see any tests for e.g. `zwave_js_server/const/command_class/thermostat.py`

firstof9:
Oops, it was a quick glance. :+1:

raman325 (approved):
thanks!

PR #342, opened 2022-01-07
Make "_value_id_from_dict" method public

raman325:
In an upcoming PR to HA I could have used this function to simplify the logic, so we may as well make it public


raman325:
this is not needed, my mistake

PR #343, opened 2022-01-07
Add functionality to record messages between lib and server

raman325:
Here's my proposal for how we will record messages. My current best idea for how to implement this in HA is:
Add a UI control somewhere that, when clicked, reloads the integration but sets `record_messages` to `True` in the client initialization. We can detect whether or not the record message functionality is enabled and display a different UI control in the same place to allow the user stop the recording and download the recording once they are done trying to repro their given issue.

As a bonus, we could offer a mechanism in the UI control to allow the user to either reload the config entry to watch everything from startup or just record from that moment onwards depending on the use case (the implementation supports both)

### DIFF
--- a/test/test_client.py
+++ b/test/test_client.py
@@ -342,7 +342,7 @@ async def test_command_error_handling(client, mock_command):
 async def test_record_messages(wallmote_central_scene, mock_command, uuid4):
     """Test recording messages."""
     client = wallmote_central_scene.client
-    assert not client.messages_are_being_recorded
+    assert not client.recording_messages
     assert not client._recorded_commands
     assert not client._recorded_events
     client.begin_recording_messages()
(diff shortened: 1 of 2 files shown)


MartinHjelmare:
Could we reuse the dump feature we already have for this purpose? It allows keeping the connection open for a timeout.

raman325:
> Could we reuse the dump feature we already have for this purpose? It allows keeping the connection open for a timeout.
> 
> https://github.com/home-assistant-libs/zwave-js-server-python/blob/7b0c9e4c5819ffc3eb3d35bbc829a546eb56d6d6/zwave_js_server/dump.py#L10-L15

You want to use the same client as the integration uses so that you can see the commands being sent and the responses coming from the server, right? I don't think it makes sense to embed this in the dump logic then

MartinHjelmare:
Yeah, I guess a second client won't see the messages we send from the first client, event if all messages sent from the server are seen by both clients.

MartinHjelmare (approved):
Looks good!

PR #346, opened 2022-01-10
Add thermostat fan `mode`, `off` and `state` property constants

EiNSTeiN-:
I noticed the `off` constant for the [`ThermostatFanMode` CC](https://github.com/zwave-js/node-zwave-js/blob/583645c6618ae1abb2666eb81806f8d73c97d719/packages/zwave-js/src/lib/commandclass/ThermostatFanModeCC.ts#L117-L125) is missing. For completeness, I also added `mode` and `state` which is a property used on the [`ThermostatFanState` CC](https://github.com/zwave-js/node-zwave-js/blob/583645c6618ae1abb2666eb81806f8d73c97d719/packages/zwave-js/src/lib/commandclass/ThermostatFanStateCC.ts#L63-L64). The `climate` platform in core uses the `THERMOSTAT_MODE_PROPERTY` and `THERMOSTAT_OPERATING_STATE_PROPERTY` which are set to the same value (mode/state), but I thought it would be better to have correctly named properties to avoid confusion when using the ThermostatFan CCs.


raman325 (approved):
thanks!

PR #354, opened 2022-01-27
Update node state data when values get added/updated/removed

raman325:
As discussed in a previous PR, this PR updates the node and value state data when we get value added/updated/removed events and makes it appear to be the same format as we get for value data when we get the first dump

### DIFF
--- a/test/model/test_node.py
+++ b/test/model/test_node.py
@@ -552,6 +552,34 @@ async def test_value_updated_events(multisensor_6):
     assert "prevValue" not in val.data
 
 
+async def test_value_removed_events(multisensor_6):
+    """Test Node value removed events."""
+    node = multisensor_6
+    event = Event(
+        type="value removed",
+        data={
+            "source": "node",
+            "event": "value removed",
+            "nodeId": 52,
+            "args": {
+                "commandClassName": "Configuration",
+                "commandClass": 112,
+                "endpoint": 0,
+                "property": 2,
+                "propertyName": "Stay Awake in Battery Mode",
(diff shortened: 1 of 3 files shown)


test/model/test_node.py
```diff
@@ -506,7 +510,77 @@ async def test_value_added_events(multisensor_6):
...
+    )
+    node.handle_value_updated(event)
+    assert isinstance(event.data["value"], ConfigurationValue)
+    assert isinstance(node.values[value_id], ConfigurationValue)
+    # ensure that the value was added to the node's state data
+    assert (value_idx := node.value_data_idx(value_id))
+    # ensure that the node's state data was updated and that old keys were removed
+    assert (value_data := node.data["values"][value_idx]) is not None
+    assert value_data["metadata"]
+    assert value_data["value"] == 1
```
MartinHjelmare, line 547: Maybe assert before the event the original value of the things we assert as updated after the event?

test/model/test_node.py
```diff
@@ -506,7 +510,77 @@ async def test_value_added_events(multisensor_6):
...
+                "endpoint": 0,
+                "property": 2,
+                "propertyName": "Stay Awake in Battery Mode",
+                "prevValue": 0,
+            },
+        },
+    )
+    node.handle_value_removed(event)
+    assert isinstance(event.data["value"], ConfigurationValue)
+    # ensure that the value was removed form the nodes value's dict
```
MartinHjelmare, line 579:
```suggestion
    # ensure that the value was removed from the nodes value's dict
```

PR #360, opened 2022-02-10
Handle unrecognized notification types

raman325:
A new CC notification was added at some point but because of the way we handled it, it created an error upstream. In this case we catch the error and report it with a more clear message.

Notification event in question: https://zwave-js.github.io/node-zwave-js/#/api/node?id=powerlevel-cc

I wanted to know, do we want to issue a patch release with this fix? If so I can target a new branch and release from that one before merging that branch into `master` (we have to do this instead of releasing from `master` so we don't force users to update the server until the next release). I can't decide since we raise an exception either way (currently the HA code raises an exception because we try to access a property that doesn't exist).

Another point not as important but worth mentioning - I'm going to add graceful handling of unsupported notification CC's in HA but I can't do that until this bug is fixed

### DIFF
--- a/zwave_js_server/exceptions.py
+++ b/zwave_js_server/exceptions.py
@@ -4,6 +4,7 @@ from typing import TYPE_CHECKING, Optional
 from .const import CommandClass
 
 if TYPE_CHECKING:
+    from .event import Event
     from .model.value import Value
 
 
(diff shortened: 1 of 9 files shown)


zwave_js_server/exceptions.py
```diff
@@ -125,7 +126,7 @@ class BulkSetConfigParameterFailed(BaseZwaveJSServerError):
     """
 
 
-class InvalidCommandClass(BaseZwaveJSServerError):
+class ValueHasInvalidCommandClass(BaseZwaveJSServerError):
```
MartinHjelmare, line 129: Do we need to rename the exception? It's a breaking change.

MartinHjelmare:
Do we have user reports affected by this problem? If there's still an error raised regardless, it doesn't seem worth a patch release from different branch.

raman325:
> Do we have user reports affected by this problem? If there's still an error raised regardless, it doesn't seem worth a patch release from different branch.

@kpine came across it but we've only had one report and to your point, nothing bad happens, there's an error, so I agree

MartinHjelmare (approved):

PR #361, opened 2022-02-10
Add strict validation via pydantic for all events

raman325:
This PR adds strict typing via `pydantic` to event data coming from the server and provides utilities for downstream clients, like HA, to leverage the strict schemas to validate event filters (e.g. the `zwave_js.event` trigger proposed in the HA core repo).

What's not being done in this PR is using the `pydantic` models internally instead of the `TypedDict` values that we store whose typing is less strict. That will come in a second PR.

A third possible PR is to replace dataclasses with `pydantic` data classes. I'm not clear yet on what benefits we get from it yet but something worth exploring


zwave_js_server/model/controller/event_model.py
```diff
@@ -0,0 +1,139 @@
...
+
+
+class ValidateDSKAndEnterPINEventModel(BaseControllerEventModel):
+    """Model for `validate dsk and enter pin` event data."""
+
+    event: Literal["validate dsk and enter pin"]
+    dsk: str
+
+
+ExclusionFailedEventModel = _event_model_factory(
```
MartinHjelmare, line 95: This is not so readable. I'd prefer defining the models as above. It's just one line difference in terms of code lines needed.

raman325: fair enough

MartinHjelmare:
We don't need Python 3.8 but I don't think we can assume Python 3.9.2, just Python 3.9.0.

raman325:
@MartinHjelmare it seems that all of the missed coverage is related to importing `TypedDict` from `typing` instead of `typing_extension`. But shouldn't our 3.8 tests hit that code path?

This PR is theoretically ready, but we should probably do some extensive testing against an actual server to make sure our event schemas make sense because validation is now running in the path of event handling. If we don't want to do that/de-risk this PR, I can remove that validation step, but the advantage of it is that we will catch mistyping in our models. I hit multiple typing bugs that I was able to fix by running the tests with event validation in place

zwave_js_server/model/controller/__init__.py
```diff
@@ -830,6 +646,10 @@ def receive_event(self, event: Event) -> None:
                 f"Controller doesn't know how to handle/forward this event: {event.data}"
             )
 
+        validate_event_data(
```
MartinHjelmare, line 649: Shouldn't we return the pydantic model instance here and use it? That's one of the big advantages afterwards that we have a typed higher level instance to use instead of a dict with base types. Is that the second step of the refactor?

raman325: Looks like we can use this function to effectively create a new model: https://github.com/samuelcolvin/pydantic/blob/master/pydantic/main.py#L610

raman325: That method doesn't validate the update data, and this issue has some more ideas on how to solve for it: https://github.com/samuelcolvin/pydantic/issues/1673

raman325:
I have been running this on my production instance for the last 24+ hours and everything seems to be OK. Because I am testing this against production, I am unable to test all the inclusion/exclusion events, but given that we have test coverage for every event type I feel pretty good about the schemas created here

zwave_js_server/model/node/event_model.py
```diff
@@ -0,0 +1,219 @@
...
+    acknowledged: int
+    total: int
+
+
+class ValueEventModel(BaseNodeEventModel):
+    """
+    Model for `value` events data.
+
+    Subclass for event models for `metadata updated`, `value added`,
+    `value notification`m `value removed`, and `value updated`.
```
MartinHjelmare, line 143:
```suggestion
    `value notification`, `value removed`, and `value updated`.
```

zwave_js_server/model/controller/inclusion_and_provisioning.py
```diff
@@ -78,7 +78,6 @@ def from_dict(cls, data: Dict[str, Any]) -> "ProvisioningEntry":
         )
 
 
-@dataclass
```
MartinHjelmare, line 71: This should be a dataclass.

raman325: it is, it's just in `controller.__init__`

raman325: ah I accidentally removed the decorator from one of the classes, thanks for catching that 👍🏾

zwave_js_server/model/controller/__init__.py
```diff
@@ -646,9 +648,12 @@ def receive_event(self, event: Event) -> None:
...
             )
 
-        validate_event_data(
-            event.data, "controller", event.type, CONTROLLER_EVENT_MODEL_MAP
-        )
+        if event.type not in CONTROLLER_EVENT_MODEL_MAP:
+            raise TypeError(f"Unknown controller event type: {event.type}")
+        try:
+            CONTROLLER_EVENT_MODEL_MAP[event.type](**event.data)
+        except ValidationError as exc:
```
MartinHjelmare, line 655: Do we even need to catch these errors? Maybe it's enough to just let them be raised as is?

zwave_js_server/model/controller/__init__.py
```diff
@@ -646,9 +648,12 @@ def receive_event(self, event: Event) -> None:
                 f"Controller doesn't know how to handle/forward this event: {event.data}"
             )
 
-        validate_event_data(
-            event.data, "controller", event.type, CONTROLLER_EVENT_MODEL_MAP
-        )
+        if event.type not in CONTROLLER_EVENT_MODEL_MAP:
```
MartinHjelmare, line 651: I think we can remove this check and let a KeyError happen if the event type is missing.

zwave_js_server/model/controller/__init__.py
raman325: I guess not!

MartinHjelmare (approved):
Nice!

PR #376, opened 2022-02-25
Fix: async_get_associations missing that a list of associations reported in a group

Dummy0815:
Now including also adapted test


MartinHjelmare:
You don't need to close the PR if you need to push more commits to fix some lint errors. Just push more commits in the same branch and the PR will be updated. This makes our review easier.

PR #381, opened 2022-03-05
Add support for zwave-js v9

raman325:
~~Still need to add tests and fix some things but using this to track my progress.~~

More details here:
- https://github.com/zwave-js/zwave-js-server/pull/541
- https://github.com/zwave-js/node-zwave-js/releases/tag/v9.0.0
- https://zwave-js.github.io/node-zwave-js/#/getting-started/migrating-to-v9


zwave_js_server/model/node/statistics.py
```diff
@@ -65,3 +82,34 @@ def commands_dropped_tx(self) -> int:
...
+    def rssi(self) -> Optional[Union[str, int]]:
+        """
+        Return average RSSI of frames received by this node.
+
+        Consecutive non-error measurements are combined using an exponential moving
+        average.
+        """
+        if (rssi := self.data.get("rssi")) is None:
+            return None
+        return friendly_rssi(rssi)
```
MartinHjelmare, line 105: I'd raise an exception with the RSSIError IntEnum as an attribute and the friendly message as exception argument, if there's an rssi error. That way we can make the typing more narrow and make the flow handling more clear for users.

raman325: The only limitation with this approach is that if a client attempts to get `RouteStatistics.repeater_rssi` and any rssi value in the list is an error, the user won't be able to see the rest of the list due to this error. Do we feel that is acceptable?

MartinHjelmare: Maybe add the complete list with values as an attribute on the exception? That way it's clear that there's an error and the library user can still access the remaining data.

raman325: Is this what you had in mind?

zwave_js_server/exceptions.py
```diff
@@ -162,11 +162,22 @@ def __init__(self, value: "Value", path: str) -> None:
...
         super().__init__(
             "Notification received with unsupported command class "
             f"{command_class.name}: {event.data}"
         )
+
+
+class RssiErrorReceived(BaseZwaveJSServerError):
+    """Exception raised when an RSSI error is received."""
+
+    def __init__(self, error: Union["RssiError", List[int]]) -> None:
```
MartinHjelmare: Let's use separate arguments and attributes for the different types to avoid ambiguity.

MartinHjelmare: Already stale.

raman325: I've switched it so that the conversion is handled in the exception itself

zwave_js_server/exceptions.py
```diff
@@ -162,11 +163,32 @@ def __init__(self, value: "Value", path: str) -> None:
...
+class RssiErrorReceived(BaseZwaveJSServerError):
+    """Exception raised when an RSSI error is received."""
+
+    def __init__(self, error: "RssiError") -> None:
+        """Initialize an RSSI error."""
+        self.error = error
+        super().__init__()
+
+
+class RssiErrorReceivedInList(BaseZwaveJSServerError):
```
MartinHjelmare, line 185: Maybe rename this to make it easier for users to know which exception to catch?
```suggestion
class RepeaterRssiErrorReceived(BaseZwaveJSServerError):
```

zwave_js_server/model/node/statistics.py
MartinHjelmare: Yes, great!

MartinHjelmare (approved):
Great!

PR #382, opened 2022-03-06
Changes in controler association functions

Dummy0815:
Fix: async_get_association_groups,
     async_get_associations,
     async_is_association_allowed,
     async_add_associations,
     async_remove_associations:
     No allows Assocation (AssociationAddress) as Source (nodeid + (optional)endpoint)
Fix: async_get_associations:
     missing that a list of associations reported in a group
Fix: async_is_association_allowed,
     async_add_associations,
     async_remove_associations:
     Now make a different between association with and without endpoint.
     As for some devices "nodeId=x endpoint=0" is not the same as "nodeid=x"

### DIFF
--- a/zwave_js_server/model/controller/__init__.py
+++ b/zwave_js_server/model/controller/__init__.py
@@ -434,7 +434,7 @@ class Controller(EventBase):
         )
         groups = {}
         for key, group in data["groups"].items():
-            groups[key] = AssociationGroup(
+            groups[int(key)] = AssociationGroup(
                 max_nodes=group["maxNodes"],
                 is_lifeline=group["isLifeline"],
                 multi_channel=group["multiChannel"],
@@ -459,7 +459,7 @@ class Controller(EventBase):
         )
         associations_map = {}
         for key, associations in data["associations"].items():
-            associations_map[key] = [
+            associations_map[int(key)] = [
                 Association(
                     node_id=association["nodeId"], endpoint=association.get("endpoint")
                 )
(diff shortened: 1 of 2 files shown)


raman325:
@Dummy0815 you need to increase the code coverage on this PR. If you click details next to the failed CI step, it will show you exactly what parts of the code aren't covered with tests.

zwave_js_server/model/controller/__init__.py
```diff
@@ -420,13 +420,16 @@ async def async_is_failed_node(self, node_id: int) -> bool:
         return cast(bool, data["failed"])
 
     async def async_get_association_groups(
-        self, node_id: int
+        self, source: Association
```
raman325, line 423: I realize you didn't create this class, but since you are cleaning up, `Association` should probably renamed to `AssociationAddress` to match the zwave-js project

zwave_js_server/model/controller/__init__.py
```diff
@@ -420,13 +420,16 @@ async def async_is_failed_node(self, node_id: int) -> bool:
...
     ) -> Dict[int, AssociationGroup]:
         """Send getAssociationGroups command to Controller."""
         data = await self.client.async_send_command(
             {
                 "command": "controller.get_association_groups",
-                "nodeId": node_id,
+                "nodeId": source.node_id,
+                **(
```
raman325, line 430: this syntax is a little hard to read. Instead of conditionally adding the key this way, can we create a payload dict and then:
```
if source.endpoint is not None:
  payload["endpoint"] = source.endpoint
```
instead?

raman325:
@Dummy0815 are you planning to update this PR so we can finish reviewing it and merge?

raman325 (approved):

Issue #387, opened 2022-03-15
LogMessageContextDataType validation


MartinHjelmare:
Looks like we need to adjust the validation in ` LogMessageContextDataType`.

```
Logger: homeassistant.components.zwave_js
Source: components/zwave_js/__init__.py:602
Integration: Z-Wave JS (documentation, issues)
First occurred: 11:41:17 AM (1 occurrences)
Last logged: 11:41:17 AM

Unexpected exception: 1 validation error for LogMessageDataType context -> type unexpected value; permitted: 'value', 'node' (type=value_error.const; given=controller; permitted=('value', 'node'))
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/components/zwave_js/__init__.py", line 602, in client_listen
    await client.listen(driver_ready)
  File "/usr/local/lib/python3.9/site-packages/zwave_js_server/client.py", line 244, in listen
    self._handle_incoming_message(data)
  File "/usr/local/lib/python3.9/site-packages/zwave_js_server/client.py", line 386, in _handle_incoming_message
    self.driver.receive_event(event)  # type: ignore
  File "/usr/local/lib/python3.9/site-packages/zwave_js_server/model/driver.py", line 83, in receive_event
    DRIVER_EVENT_MODEL_MAP[event.type](**event.data)
  File "/usr/local/lib/python3.9/site-packages/pydantic/main.py", line 331, in __init__
    raise validation_error
pydantic.error_wrappers.ValidationError: 1 validation error for LogMessageDataType
context -> type
  unexpected value; permitted: 'value', 'node' (type=value_error.const; given=controller; permitted=('value', 'node'))
```

We probably need to add `"controller"` to the allowed items here:

https://github.com/home-assistant-libs/zwave-js-server-python/blob/ba38678977e5883fb339a64048ca7e7572d21c8e/zwave_js_server/model/log_message.py#L16

PR #388, opened 2022-03-16
Fix LogMessageContextDataType typing

raman325:
fixes https://github.com/home-assistant-libs/zwave-js-server-python/issues/387


MartinHjelmare (approved):
Thanks!

PR #393, opened 2022-03-27
Fix release drafter breaking change label

thecode:
Align label with core `breaking-change` label


raman325 (approved):
Thanks @thecode !!

Issue #396, opened 2022-03-29
Add support for MultilevelSwitchCommandClass 


reubenbijl:
Using node-z-wave-js beta, when receiving a multilevelSwitchCC an error is generated.

See change with Z-Wave-js - https://github.com/zwave-js/node-zwave-js/pull/4282

ZwaveJS and jwaveJSMQTT logs
19:47:35.664 DRIVER « [Node 030] [REQ] [ApplicationCommand]
                      └─[MultilevelSwitchCCStopLevelChange]
19:47:35.664 CNTRLR   [Node 030] treating MultilevelSwitchCC::StopLevelChange as a notification
2022-03-29 19:47:35.664 INFO ZWAVE: Node 30 CC Multilevel Switch { eventType: 5 }
2022-03-29 19:47:35.969 INFO ZWAVE-SERVER: Client disconnected

Home Assistant Logs
[homeassistant.components.zwave_js] Failed to listen: Notification received with unsupported command class SWITCH_MULTILEVEL: {'source': 'node', 'event': 'notification', 'nodeId': 30, 'ccId': 38, 'args': {'eventType': 5}}

PR #397, opened 2022-03-29
Add support for Multilevel Switch Notification

reubenbijl:
https://github.com/home-assistant-libs/zwave-js-server-python/issues/396


MartinHjelmare:
This requires node-zwave-js version 9, right? This PR then probably should be merged after we merge https://github.com/home-assistant-libs/zwave-js-server-python/pull/381.

reubenbijl:
Yes, correct, requires z-wave-js version 9.

zwave_js_server/model/notification.py
```diff
@@ -184,3 +184,47 @@ def status(self) -> PowerLevelTestStatus:
     def acknowledged_frames(self) -> int:
         """Return acknowledged frames property."""
         return self.data["args"]["acknowledgedFrames"]
+
+class MultilevelSwitchNotificationArgsDataType(TypedDict):
+    """Represent args for a Multi Level Switch CC notification event data dict type."""
+
+    event_type: int
+    direction: str
```
raman325, line 192:
```suggestion
class MultilevelSwitchNotificationArgsDataType(TypedDict, total=False):
    """Represent args for a Multi Level Switch CC notification event data dict type."""

    event_type: int  # required
    direction: str
```
I believe `direction` is optional here right?

reubenbijl: Yes, if the event type is a Multilevel Switch Stop event, then there is not direction associated with it.

raman325:
please run black on the file to remove the extra whitespaces

zwave_js_server/model/notification.py
```diff
@@ -184,3 +184,47 @@ def status(self) -> PowerLevelTestStatus:
...
+        """Return node ID property."""
+        return self.data["nodeId"]
+        
+    @property
+    def command_class(self) -> int:
+        """Return command class."""
+        return self.data["ccId"]
+
+    @property
+    def event_type(self) -> int:
```
raman325, line 220: We should create an `IntEnum` for `MultilevelSwitchCommand` in `zwave_js_server.const.command_class.multilevel_switch` and then return the enum value here instead of the raw int

reubenbijl: Do you mean for the event type?
ie., new enum in         return CommandClass.SWITCH_MULTILEVEL with something like this:

class MultilevelSwitchCommand(IntEnum):
	START_LEVEL = 4
	STOP = 5

Or for the Command classId?
return CommandClass.SWITCH_MULTILEVEL

reubenbijl: @raman325  see commit, I think this is what you meant.

zwave_js_server/const/command_class/multilevel_switch.py
```diff
@@ -6,3 +6,10 @@
 COVER_CLOSE_PROPERTY = "Close"
 COVER_DOWN_PROPERTY = "Down"
 COVER_OFF_PROPERTY = "Off"
+
+
+class MultilevelSwitchCommand(IntEnum):
+    """Enum for known multilevel switch notifications"""
+
+    START_LEVEL_CHANGE = 4
```
raman325, line 14: not necessary for this PR but we may as well define all the possible enums here

raman325 (approved):
this can be merged as soon as tests pass and after https://github.com/home-assistant-libs/zwave-js-server-python/pull/381 gets merged. This will happen soon, just waiting a bit to make sure there are no issues with the latest HA release before we bump the schema in that PR

raman325:
please add tests for this new notification type

raman325 (changes requested):
please add tests to improve coverage

raman325 (changes requested):
one more run of black and we **should** be good

raman325 (approved):
great, just need to merge the other PR!

PR #405, opened 2022-04-21
Log a warning instead of raising for unrecognized notification CCs

raman325:
raising an exception currently raises an exception which causes the HA integration to disconnect. This is the wrong way to handle this, we should just log a warning and let the consumer figure out how to handle this scenario

### DIFF
--- a/zwave_js_server/model/node/__init__.py
+++ b/zwave_js_server/model/node/__init__.py
@@ -792,7 +792,9 @@ class Node(EventBase):
                 self, cast(PowerLevelNotificationDataType, event.data)
             )
         else:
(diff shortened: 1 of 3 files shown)


zwave_js_server/model/node/__init__.py
```diff
@@ -58,6 +58,9 @@
     from ...client import Client
 
 
+_LOGGER = logging.getLogger(__name__)
```
MartinHjelmare, line 61:
```suggestion
_LOGGER = logging.getLogger(__package__)
```

MartinHjelmare (approved):

PR #422, opened 2022-05-16
Switch RouteStatistics.route_failed_between from list to tuple

raman325:
`routeFailedBetween` will always be a list of length two so a `tuple` is a better representation

### DIFF
--- a/zwave_js_server/model/statistics.py
+++ b/zwave_js_server/model/statistics.py
@@ -85,7 +85,9 @@ class RouteStatistics:
             return None
         assert self.client.driver
         assert len(node_ids) == 2
-        return (self.client.driver.controller.nodes[node_id] for node_id in node_ids)
+        return tuple(
+            self.client.driver.controller.nodes[node_id] for node_id in node_ids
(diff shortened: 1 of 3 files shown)


zwave_js_server/model/statistics.py
```diff
@@ -79,12 +79,15 @@ def repeater_rssi(self) -> List[int]:
         return repeater_rssi
 
     @property
-    def route_failed_between(self) -> Optional[List["Node"]]:
+    def route_failed_between(self) -> Optional[Tuple["Node", ...]]:
```
MartinHjelmare, line 82: Should we specify that it's always a two item tuple?

raman325: I tried that but with list comprehension mypy didn't like it. Should I be explicit with the indexes?

MartinHjelmare: I think we should be explicit with the typing since we know the length. The code should follow.

MartinHjelmare (approved):

PR #424, opened 2022-05-17
Get initial node statistics from state dump

raman325:


zwave_js_server/model/node/statistics.py
```diff
@@ -34,16 +34,18 @@ class NodeStatistics:
     """Represent a node statistics update."""
 
     def __init__(
-        self, client: "Client", data: Optional[NodeStatisticsDataType] = None
+        self, client: "Client", data: Optional[NodeStatisticsDataType]
     ) -> None:
         """Initialize node statistics."""
-        self.data = data or NodeStatisticsDataType(
+        self.data = NodeStatisticsDataType(
```
MartinHjelmare, line 40: Why do we change this? The former code potentially saves a call.

PR #434, opened 2022-05-27
Pass Node instead of node ID into controller node commands

raman325:
In each of these commands, the node ID referenced will be from a node already included in the model, so why not let the called pass the node in instead of the node ID?

### DIFF
--- a/test/model/test_controller.py
+++ b/test/model/test_controller.py
@@ -649,53 +649,52 @@ async def test_hash(controller):
     assert hash(controller) == hash(controller.home_id)
 
 
-async def test_remove_failed_node(controller, uuid4, mock_command):
+async def test_remove_failed_node(controller, multisensor_6, uuid4, mock_command):
     """Test remove failed node."""
     ack_commands = mock_command(
         {"command": "controller.remove_failed_node"},
         {},
     )
 
-    node_id = 52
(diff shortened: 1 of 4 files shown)


MartinHjelmare:
Are these the only cases where we can pass model instances instead of model ids? I think we should make sure we're consistent across the library before changing this.

raman325:
good question. I found some others via searching for `id` in function params. There are two left:

1. `controller.async_unprovision_smart_start_node` - This one accepts a DSK or node ID, but because these entries are created before a node is included, I think it makes sense to still accept a node ID because the node may not be an existing node.
https://github.com/home-assistant-libs/zwave-js-server-python/blob/d2566fc5080e9034a57668decc132cc2d375929d/zwave_js_server/model/controller/__init__.py#L267
2. `Node.value_data_idx` - this is an internal command used to search for a value dict in the nodes state data. I think it makes sense to leave `value_id` as an input here too.
https://github.com/home-assistant-libs/zwave-js-server-python/blob/d2566fc5080e9034a57668decc132cc2d375929d/zwave_js_server/model/node/__init__.py#L724

Those are the only two places remaining I could find that accept a potential model ID as an input rather than the model instance, but I think in both of those cases we can't guarantee there is a model instance

MartinHjelmare (approved):
Ok, good!

PR #435, opened 2022-05-29
Mypy performance fix

cdce8p:
Using a Union TypeAlias with lots of members can significantly slow down mypy, especially the uncompiled version. Noticed it while testing a mypy dev build with Home Assistant. The CI run took significantly longer, `30+ min` compared to `7-8min` normally.

The slowdown in HA only happened after the `0.37.0` update, since that's when `py.typed` was added. Previously mypy just ignored the dependency.

--
Specifically, the `MultilevelSensorScaleType` alias with 30+ members is a problem. The fix here is relatively simple. We can create a common enum base class and use that instead of the union TypeAlias. The only thing to remember is that no members can be added to the base class directly. Otherwise subclassing enums wouldn't work.

### DIFF
--- a/scripts/generate_multilevel_sensor_constants.py
+++ b/scripts/generate_multilevel_sensor_constants.py
@@ -114,10 +114,11 @@ def generate_int_enum_class_definition(
     enum_ref_url: str | None = None,
     get_id_func: Callable | None = None,
     docstring_info: str = "",
+    base_class: str = "IntEnum",
 ) -> List[str]:
     """Generate an IntEnum class definition as an array of lines of string."""
     class_def = []
-    class_def.append(f"class {class_name}(IntEnum):")
+    class_def.append(f"class {class_name}({base_class}):")
     docstring = (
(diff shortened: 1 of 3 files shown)


scripts/generate_multilevel_sensor_constants.py
```diff
@@ -133,6 +134,14 @@ def generate_int_enum_class_definition(
     return class_def
 
 
+def generate_int_enum_base_class(class_name: str, docstring: str):
```
MartinHjelmare, line 137:
```suggestion
def generate_int_enum_base_class(class_name: str, docstring: str) -> List[str]:
```

MartinHjelmare (approved):
Looks good! Thanks!

cdce8p:
Small followup now that Home Assistant has been updated to use `0.37.1`, even the normal mypy runs are significantly faster. Sub `1:45min` for a full mypy CI run, compared to `2:30-3min` previously. Wasn't expecting it, but obviously great to see 🚀

raman325:
excellent, thanks for your help with this!

PR #440, opened 2022-06-08
Clean up tests

raman325:
~~It turns out that using the `Client` class to create a new connection and perform a firmware update causes problems. I can't explain why, but when using this approach, after dispatching the command we don't ever receive the response. To fix this I mirrored the dump logic and was able to successfully get a response to HA from a firmware update.~~ EDIT: The problem listed here is fixed in a better way in #441. This PR has been changed to clean up tests.


zwave_js_server/firmware.py
```diff
@@ -17,18 +19,46 @@ async def begin_firmware_update(
     file_format: Optional[str] = None,
 ) -> None:
     """Send beginFirmwareUpdate command to Node."""
-    client = Client(url, session)
-    await client.connect()
-    await client.set_api_schema()
+    client = await session.ws_connect(url, compress=15, max_msg_size=0)
```
MartinHjelmare, line 22: We're missing some exception handling and error raising here and around the received message and for invalid server version when not using the client.

MartinHjelmare (approved):
Nice!

PR #441, opened 2022-06-08
Fix firmware update logic

raman325:
This PR supersedes https://github.com/home-assistant-libs/zwave-js-server-python/pull/440 which will switch to a test refactor.

The reason we weren't receiving messages before is because the loop to receive connections wasn't running. I broke that out into a separate function so now we can start the loop in the firmware upload logic. This requires no change to tests while eliminating repeat logic

### DIFF
--- a/zwave_js_server/firmware.py
+++ b/zwave_js_server/firmware.py
@@ -22,7 +22,7 @@ async def begin_firmware_update(
     await client.connect()
     await client.set_api_schema()
 
-    asyncio.get_running_loop().create_task(client.receive_until_closed())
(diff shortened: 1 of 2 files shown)


zwave_js_server/firmware.py
```diff
@@ -21,6 +22,8 @@ async def begin_firmware_update(
     await client.connect()
     await client.set_api_schema()
 
+    asyncio.get_running_loop().create_task(client.receive_until_closed())
```
raman325, line 25: do we need to ensure this task is done/cancelled after disconnect?

MartinHjelmare (approved):
Looks good!

PR #443, opened 2022-06-08
Initialize firmware progress when an update has been sent

raman325:
Right now, the lib only knows that a firmware update is in progress if we start receiving `firmware update progress` messages. There is a scenario where we attempt to update the firmware on a sleeping device - in this case, the controller will have accepted the upload, but we won't start receiving events until the device wakes up and the upload process starts. So in this change, we initialize the `firmware_update_progress` property of the node to indicate that the firmware is effectively queued.

Should we somehow handle this in the server instead? Right now we don't maintain state information about the network or its nodes in the server, but if the client disconnects, then it loses state, whereas if the server stops, then the queue and any occurring processes get terminated so the state will be correct when it restarts.

### DIFF
--- a/test/test_firmware.py
+++ b/test/test_firmware.py
@@ -28,6 +28,10 @@ async def test_begin_firmware_update_guess_format(url, client_session, multisens
             require_schema=5,
         )
         disconnect_mock.assert_called_once()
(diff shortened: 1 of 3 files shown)


raman325:
I don't know that this is the right call so I will have to think on this some more

PR #445, opened 2022-06-10
Add support for node.get_firmware_update_progress command

raman325:
The server now keeps track of when a firmware update is in progress. This is necessary for the frontend work to support firmware updates

### DIFF
--- a/zwave_js_server/model/node/__init__.py
+++ b/zwave_js_server/model/node/__init__.py
@@ -644,6 +644,14 @@ class Node(EventBase):
         )
         self.data["location"] = location
 
+    async def async_get_firmware_update_progress(self) -> bool:
+        """Send getFirmwareUpdateProgress command to Node."""
+        data = await self.async_send_command(
+            "get_firmware_update_progress", require_schema=18, wait_for_result=True
+        )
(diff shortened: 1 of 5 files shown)


zwave_js_server/model/node/__init__.py
```diff
@@ -644,6 +644,14 @@ async def async_set_location(
         )
         self.data["location"] = location
 
+    async def async_get_firmware_update_progress(self) -> bool:
+        """Send getFirmwareUpdateProgress command to Node."""
+        data = await self.async_send_command(
+            "get_firmware_update_progress", require_schema=18, wait_for_result=True
+        )
+        assert data
+        return cast(bool, data["progress"])
```
MartinHjelmare, line 653: Does a `True` return value mean that a firmware update is in progress, and `False` means there's no firmware update in progress?

raman325: That's correct. Should I add that to the docstring?

MartinHjelmare: Yes, I think it's needed.

MartinHjelmare (approved):

PR #446, opened 2022-06-11
Add support for zwave-js 9.4

raman325:
See here for more details: https://github.com/zwave-js/zwave-js-server/pull/645

I also added a new property to the `node` that lets the caller figure out whether a node needs to be interviewed manually

### DIFF
--- a/zwave_js_server/const/__init__.py
+++ b/zwave_js_server/const/__init__.py
@@ -3,14 +3,15 @@ import sys
 from enum import Enum, IntEnum
 
 # minimal server schema version we can handle
-MIN_SERVER_SCHEMA_VERSION = 17
+MIN_SERVER_SCHEMA_VERSION = 18
 # max server schema version we can handle (and our code is compatible with)
-MAX_SERVER_SCHEMA_VERSION = 17
+MAX_SERVER_SCHEMA_VERSION = 18
 
 TYPING_EXTENSION_FOR_TYPEDDICT_REQUIRED = sys.version_info < (3, 9, 2)
 
 VALUE_UNKNOWN = "unknown"
 
+NOT_INTERVIEWED = "None"
 INTERVIEW_FAILED = "Failed"
 
 CURRENT_STATE_PROPERTY = "currentState"
--- a/zwave_js_server/model/controller/__init__.py
+++ b/zwave_js_server/model/controller/__init__.py
@@ -10,7 +10,6 @@ from ...const import (
     RFRegion,
     ZwaveFeature,
 )
-from ...exceptions import InvalidCommand
 from ...event import Event, EventBase
 from ...util.helpers import convert_base64_to_bytes, convert_bytes_to_base64
 from ..association import AssociationAddress, AssociationGroup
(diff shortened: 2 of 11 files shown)


zwave_js_server/model/controller/__init__.py
```diff
@@ -691,6 +692,17 @@ async def async_get_known_lifeline_routes(
             for node_id, lifeline_routes in data["routes"].items()
         }
 
+    async def async_interview_node(self, node: Node) -> None:
```
MartinHjelmare, line 695: Z-Wave JS puts this command on the driver. Why does it do that and we put it on the controller?

Also, it says that it's not recommended to await the command. Should we use the other method to send the command, so we don't wait for a result?

https://zwave-js.github.io/node-zwave-js/#/api/driver?id=interviewnode

zwave_js_server/const/__init__.py
```diff
@@ -3,14 +3,15 @@
...
+MIN_SERVER_SCHEMA_VERSION = 18
 # max server schema version we can handle (and our code is compatible with)
-MAX_SERVER_SCHEMA_VERSION = 17
+MAX_SERVER_SCHEMA_VERSION = 18
 
 TYPING_EXTENSION_FOR_TYPEDDICT_REQUIRED = sys.version_info < (3, 9, 2)
 
 VALUE_UNKNOWN = "unknown"
 
+NOT_INTERVIEWED = "None"
```
MartinHjelmare, line 14: Should we create an enum with the possible interview stages?

https://zwave-js.github.io/node-zwave-js/#/api/node?id=interviewstage

zwave_js_server/model/controller/__init__.py
raman325: whoops, that was a mistake, thanks for catching that!

raman325: in the server, we don't await the command, we just return a response. The purpose of waiting is to make sure that we receive a response to confirm that the command was issued

zwave_js_server/const/__init__.py
raman325: When I last spoke to Al about this, he suggested not to because the interview stages were subject to change at any time (they are not part of the standard I don't think). I think `None` and `Failed` are safe to remain which is why I was comfortable creating constants for them

zwave_js_server/model/controller/__init__.py
MartinHjelmare: Ok.

zwave_js_server/const/__init__.py
MartinHjelmare: Ok.

zwave_js_server/model/driver.py
```diff
@@ -166,6 +168,16 @@ async def async_enable_error_reporting(self) -> None:
         """Send command to enable Sentry error reporting."""
         await self._async_send_command("enable_error_reporting", require_schema=16)
 
+    async def async_interview_node(self, node: Node) -> None:
+        """Send interviewNode command to Controller."""
```
MartinHjelmare:
```suggestion
        """Send an interviewNode command."""
```

MartinHjelmare (approved):

MartinHjelmare (approved):

PR #452, opened 2022-06-18
Add missing node.get_firmware_update_capabilities command

raman325:
This command has been available for a while but we haven't taken advantage of it.

### DIFF
--- a/zwave_js_server/model/firmware.py
+++ b/zwave_js_server/model/firmware.py
@@ -36,7 +36,7 @@ class FirmwareUpdateCapabilities:
         return self.data["firmwareUpgradable"]
 
     @property
-    def firmware_targets(self) -> Optional[List[int]]:
+    def firmware_targets(self) -> List[int]:
         """Return firmware targets."""
         if not self.firmware_upgradable:
             raise TypeError("Firmware is not upgradeable.")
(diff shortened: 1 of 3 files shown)


zwave_js_server/model/firmware.py
```diff
@@ -13,6 +13,67 @@
...
+    def supports_activation(self) -> Optional[bool]:
+        """Return whether node supports delayed activation of the new firmware."""
+        if not self.firmware_upgradable:
+            raise TypeError("Firmware is not upgradeable.")
+        if (val := self.data["supportsActivation"]) == VALUE_UNKNOWN:
+            return None
+        assert isinstance(val, bool)
+        return val
+
+    def to_dict(self) -> Dict[str, Optional[Union[bool, List[int]]]]:
```
MartinHjelmare, line 65: Let's return a TypedDict here.

zwave_js_server/model/firmware.py
```diff
@@ -13,6 +13,67 @@
...
+    continuesToFunction: Union[bool, str]
+    supportsActivation: Union[bool, str]
+
+
+class FirmwareUpdateCapabilities:
+    """Model for firmware update capabilities."""
+
+    def __init__(self, node: "Node", data: FirmwareUpdateCapabilitiesDataType) -> None:
+        """Initialize class."""
+        self.node = node
```
MartinHjelmare, line 30: Why do we set the node here? It's not used.

zwave_js_server/model/firmware.py
```diff
@@ -13,6 +13,67 @@
...
+class FirmwareUpdateCapabilitiesDataType(TypedDict, total=False):
+    """Represent a firmware update capabilities dict type."""
+
+    firmwareUpgradable: bool  # required
+    firmwareTargets: List[int]
+    continuesToFunction: Union[bool, str]
+    supportsActivation: Union[bool, str]
+
+
+class FirmwareUpdateCapabilities:
```
MartinHjelmare, line 25: We could describe the response with a pydantic model directly.

zwave_js_server/model/firmware.py
raman325: I thought it was needed when I first started but you are right, I'll remove

zwave_js_server/model/firmware.py
raman325: if we are going to do that should we be looking at that elsewhere as well? I almost feel like it should be done as a single big initiative like we did with the event models, what do you think?

MartinHjelmare: We should probably do it in the same release, yes. We don't need to do everything in the same PR, I think.

MartinHjelmare (approved):

zwave_js_server/model/firmware.py
raman325: Good point

PR #453, opened 2022-06-19
Add support for providing target for firmware

raman325:
related to https://github.com/zwave-js/zwave-js-server/pull/653

### DIFF
--- a/test/test_firmware.py
+++ b/test/test_firmware.py
@@ -30,7 +30,9 @@ async def test_begin_firmware_update_guess_format(url, client_session, multisens
         disconnect_mock.assert_called_once()
 
 
-async def test_begin_firmware_update_known_format_and_target(url, client_session, multisensor_6):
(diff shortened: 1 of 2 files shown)


test/test_firmware.py
```diff
@@ -41,7 +43,7 @@ async def test_begin_firmware_update_known_format(url, client_session, multisens
     ) as disconnect_mock:
         node = multisensor_6
         await begin_firmware_update(
-            url, node, "test", bytes(10), client_session, "test"
+            url, node, "test", bytes(10), client_session, "test", 0
```
MartinHjelmare, line 46: Side note: It's easier to read if we use keyword arguments when there are many arguments.

MartinHjelmare (approved):

test/test_firmware.py
raman325: noted

PR #457, opened 2022-06-21
Bump zwave-js-server-python to 0.39.0

raman325:
It's a minor bump because I find it helpful for the versions (both server and lib) to map well to the schema version. let me know if you disagree

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -5,7 +5,7 @@ from setuptools import find_packages, setup
 
 PROJECT_DIR = Path(__file__).parent.resolve()
 README_FILE = PROJECT_DIR / "README.md"
-VERSION = "0.38.0"
+VERSION = "0.39.0"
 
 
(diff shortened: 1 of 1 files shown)


MartinHjelmare (approved):

raman325:
@MartinHjelmare  I guess we should keep this to 0.39.0 instead of skipping to 0.40.0 even though we are skipping straight to schema 20 huh?

MartinHjelmare (approved):
We could introduce a version bump schema that keeps some identifiers in sync with the server. But then I think we should first document that.

raman325:
> We could introduce a version bump schema that keeps some identifiers in sync with the server. But then I think we should first document that.

Agreed. Seems simple though - we can just keep major and minor versions in sync, and say that minor version is the max schema number currently supported

PR #480, opened 2022-09-06
Fix value updated logic

raman325:
it appears that we have had a bug for some time. Some devices will include the `value` parameter in the `value updated` event. With the current logic, we would use the `value` data (which is basically `prevValue`) and overwrite `newValue`. The reason this didn't get caught before is because after a value update, a refresh would be performed on non supervised CCs that would update the state to the correct value. With v10 and the introduction of supervision to more CCs, this refresh doesn't occur, exposing the bug.

You can confirm this bug by removing the change to the `Value.update` method and keeping the changes to the tests.

Thanks to @kpine for finding this one

### DIFF
--- a/zwave_js_server/model/controller/inclusion_and_provisioning.py
+++ b/zwave_js_server/model/controller/inclusion_and_provisioning.py
@@ -151,6 +151,7 @@ class QRProvisioningInformation(ProvisioningEntry, QRProvisioningInformationMixi
     @classmethod
     def from_dict(cls, data: Dict[str, Any]) -> "QRProvisioningInformation":
         """Return QRProvisioningInformation from data dict."""
+        # pylint: disable=unexpected-keyword-arg
         cls_instance = cls(
             version=QRCodeVersion(data["version"]),
             security_classes=[
--- a/test/model/test_node.py
+++ b/test/model/test_node.py
@@ -568,6 +568,7 @@ async def test_value_updated_events(multisensor_6):
                 "endpoint": 0,
                 "property": 2,
                 "propertyName": "Stay Awake in Battery Mode",
+                "value": -1,
                 "newValue": 1,
                 "prevValue": 0,
             },
@@ -587,6 +588,7 @@ async def test_value_updated_events(multisensor_6):
     # ensure that the value's state data was updated and that old keys were removed
     val = node.values[value_id]
     assert val.data["value"] == 1
+    assert val.value == 1
     assert "newValue" not in val.data
     assert "prevValue" not in val.data
 
--- a/zwave_js_server/model/value.py
+++ b/zwave_js_server/model/value.py
@@ -265,7 +265,7 @@ class Value:
         self.data.update(data)
         if "newValue" in data:
             self._value = self.data["value"] = data["newValue"]
-        if "value" in data:
+        elif "value" in data:
             self._value = self.data["value"] = data["value"]
         if "metadata" in data:
             self._metadata.update(data["metadata"])


zwave_js_server/model/controller/inclusion_and_provisioning.py
```diff
@@ -151,6 +151,7 @@ def to_dict(self) -> Dict[str, Any]:
     @classmethod
     def from_dict(cls, data: Dict[str, Any]) -> "QRProvisioningInformation":
         """Return QRProvisioningInformation from data dict."""
+        # pylint: disable=unexpected-keyword-arg
```
MartinHjelmare, line 154: Why do we need the disable?

raman325: the new pylint version triggers an error for each of the keyword arguments. We use keyword args for readability because it's such a big list.

raman325: I think the right long term fix is to force keyword arguments everywhere and we can remove the disable. I'll log an issue to track that

AlCalzone:
I'm not sure where the `value` prop comes from. Is zwave-js-server setting it?

raman325:
> I'm not sure where the `value` prop comes from. Is zwave-js-server setting it?

It appears to be coming from the driver because we just pass the event on as is in the server:
https://github.com/zwave-js/zwave-js-server/blob/master/src/lib/forward.ts#L200-L214

AlCalzone:
freshcoast figured it out. That comes from HA originally as part of an overly-filled value ID. The driver doesn't filter out the unnecessary properties.

zwave_js_server/model/controller/inclusion_and_provisioning.py
MartinHjelmare: Sounds like a pylint bug.

MartinHjelmare (approved):

kpine:
This is what a set_value message looks like published from HA:

```text
2022-09-06 09:34:58.786 DEBUG (MainThread) [zwave_js_server] Publishing message:
{'command': 'node.set_value',
 'messageId': '4d9a1e01bbcd401f85aef74efeabba7d',
 'nodeId': 11,
 'value': 0,
 'valueId': {'ccVersion': 2,
             'commandClass': 64,
             'commandClassName': 'Thermostat Mode',
             'endpoint': 0,
             'metadata': {'label': 'Thermostat mode',
                          'max': 255,
                          'min': 0,
                          'readable': True,
                          'states': {'0': 'Off',
                                     '1': 'Heat',
                                     '11': 'Energy heat',
                                     '12': 'Energy cool',
                                     '2': 'Cool',
                                     '3': 'Auto'},
                          'type': 'number',
                          'writeable': True},
             'property': 'mode',
             'propertyName': 'mode',
             'value': 1}}
```

The `valueId` is not really correct, according to the [docs](https://zwave-js.github.io/node-zwave-js/#/api/valueid?id=valueid), although it accepts it. I think this is what the above message should ideally look like:

```
2022-09-06 09:34:58.786 DEBUG (MainThread) [zwave_js_server] Publishing message:
{'command': 'node.set_value',
 'messageId': '4d9a1e01bbcd401f85aef74efeabba7d',
 'nodeId': 11,
 'value': 0,
 'valueId': {'ccVersion': 2,
             'commandClass': 64,
             'endpoint': 0,
             'property': 'mode'}}
```

So this is sort of a self-inflicted bug, and the solution here, while it works, is also just coding around the self-infliction. There should never even be a `value` field in a Value event that would need to be checked for.

AlCalzone (changes requested):
What @kpine said :)

raman325:
OK I am going to merge this to stop the bleeding and then we can fix that other issue separately

Theflxkid:
After updating HA specifically to get this fix (which it appears to have fixed) the UI to actually change the temperature disappeared. Any chance it's related to this fix? I've cleared cache and tried in different browsers, etc.

Here is a screenshot of what my thermostat control looks like (missing actual control to change set temperature)

image

PR #482, opened 2022-09-07
Filter value ID dict data

raman325:
The reason https://github.com/home-assistant-libs/zwave-js-server-python/pull/480 became an issue is because we were previously sending the full value data, including metadata and the value of the value, to the driver. The driver would store this data in the value DB and then send it back but the interface expects just the four properties (cc, endpoint, property, property_key). In this PR we filter the data down to only send the dict value.

Note that I changed `get_value_id` to `get_value_id_str` to distinguish between the dict value ID and the string value ID - this will have to be updated in core when we bump the dependency to include this PR


MartinHjelmare (approved):
Thanks!

PR #487, opened 2022-09-08
Drop python 3.8 support

raman325:
The minimum for HA is now python 3.9 so I think we can drop 3.8 support.


raman325:
any idea why we are getting this failure? I can't reproduce it locally...

MartinHjelmare (approved):

PR #494, opened 2022-09-15
`async_get_state` cmds return state instead of updating it

raman325:
It occurred to me while troubleshooting an issue that it would be nice to get both the lib state and driver state in the device diagnostics dump (config entry gets the driver state, we may want to consider adding the lib state). We have the async_get_state command but currently it just updates the internal state. With this new flag, an application like HA could call this command to retrieve the state without updating the internal state for diagnostics purposes

### DIFF
--- a/zwave_js_server/model/node/__init__.py
+++ b/zwave_js_server/model/node/__init__.py
@@ -660,14 +660,12 @@ class Node(EventBase):
         assert data
         return RouteHealthCheckSummary(data["summary"])
 
-    async def async_get_state(self, update_internal_state: bool = True) -> NodeDataType:
+    async def async_get_state(self) -> NodeDataType:
         """Get node state."""
         data = await self.async_send_command(
             "get_state", require_schema=14, wait_for_result=True
         )
         assert data
-        if update_internal_state:
-            self.update(data["state"])
         return cast(NodeDataType, data["state"])
 
     async def async_set_name(
--- a/zwave_js_server/model/node/__init__.py
+++ b/zwave_js_server/model/node/__init__.py
@@ -660,7 +660,7 @@ class Node(EventBase):
         assert data
         return RouteHealthCheckSummary(data["summary"])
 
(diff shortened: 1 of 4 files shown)


zwave_js_server/model/node/__init__.py
```diff
@@ -660,13 +660,15 @@ async def async_check_route_health(
...
 
-    async def async_get_state(self) -> None:
+    async def async_get_state(self, update_internal_state: bool = True) -> NodeDataType:
         """Get node state."""
         data = await self.async_send_command(
             "get_state", require_schema=14, wait_for_result=True
         )
         assert data
-        self.update(data["state"])
+        if update_internal_state:
```
MartinHjelmare, line 669: Why shouldn't we always update the internal state? The retrieved state is correct, right?

raman325: if we are getting the state from the driver for diagnostics, we don't want to update the internal state as it may cause the user to think that a problem they are trying to troubleshoot has been resolved when it hasn't been

raman325: in most cases we will keep the default `update_internal_state == True`. We would only set it to false as part of device diagnostics

raman325: I guess we could use the new get_node_unsafe command which effectively does the same thing without having to modify this command. Thoughts?

MartinHjelmare: Maybe we should remove the `update` call from this method and the library user can update the node state when needed instead after getting the node state?

raman325: yeah we could do that. In HA this is only called in one place at the moment

MartinHjelmare: I think that is more explicit and composable.

MartinHjelmare (approved):
Thanks!

PR #495, opened 2022-09-15
Allow upstream application to provide its user agent

raman325:
Implements https://github.com/zwave-js/node-zwave-js/pull/5070 which hasn't been released yet and doesn't have server support but will soon


MartinHjelmare (approved):

raman325:
should the lib always add itself to the user agent? While it may be redundant for HA for the most part, I don't see a downside to doing this, only an upside

MartinHjelmare (approved):

PR #497, opened 2022-09-17
Improve InvalidServerVersion exception and corresponding message

raman325:
We have no structured way to get the current version and required versions right now so I added that. I also think our connection error message was a little vague so I updated it to match the async_send_command require_schema exception message which seems more clear to me.

### DIFF
--- a/zwave_js_server/exceptions.py
+++ b/zwave_js_server/exceptions.py
@@ -64,11 +64,16 @@ class InvalidServerVersion(BaseZwaveJSServerError):
     """Exception raised when connected to server with incompatible version."""
 
     def __init__(
-        self, current_version: int, required_version: int, message: str
+        self,
+        server_version: str,
+        server_max_schema_version: int,
+        required_schema_version: int,
+        message: str,
     ) -> None:
         """Initialize an invalid server version error."""
-        self.current_version = current_version
-        self.required_version = required_version
+        self.server_version = server_version
+        self.server_max_schema_version = server_max_schema_version
+        self.required_schema_version = required_schema_version
         super().__init__(message)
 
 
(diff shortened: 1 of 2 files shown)


zwave_js_server/exceptions.py
```diff
@@ -63,6 +63,12 @@ class InvalidMessage(BaseZwaveJSServerError):
 class InvalidServerVersion(BaseZwaveJSServerError):
     """Exception raised when connected to server with incompatible version."""
 
+    def __init__(self, current_version: int, required_version: int, message: str) -> None:
+        """Initialize a invalid server version error."""
```
MartinHjelmare, line 67:
```suggestion
        """Initialize an invalid server version error."""
```

MartinHjelmare (approved):

zwave_js_server/exceptions.py
```diff
@@ -63,6 +63,14 @@ class InvalidMessage(BaseZwaveJSServerError):
 class InvalidServerVersion(BaseZwaveJSServerError):
     """Exception raised when connected to server with incompatible version."""
 
+    def __init__(
+        self, current_version: int, required_version: int, message: str
```
MartinHjelmare, line 67:
```suggestion
        self, current_version: str, required_version: int, message: str
```

MartinHjelmare (commented):
It may be a bit weird that the current version and the required version are different types.

zwave_js_server/exceptions.py
raman325: looked into it, current_version is the server version, required_version is the schema version. Changes incoming

MartinHjelmare (approved):

zwave_js_server/exceptions.py
```diff
@@ -63,6 +64,19 @@ class InvalidMessage(BaseZwaveJSServerError):
...
     """Exception raised when connected to server with incompatible version."""
 
+    def __init__(
+        self,
+        version_info: Optional["VersionInfo"],
+        required_schema_version: int,
+        message: str,
+    ) -> None:
+        """Initialize an invalid server version error."""
+        assert version_info
```
MartinHjelmare, line 74: We should not accept an optional version if that's not ok. Make the callers check that they pass the correct argument instead.

PR #560, opened 2023-01-24
Add support for zwave-js-server 1.25.0

raman325:
See https://github.com/zwave-js/node-zwave-js/releases/tag/v10.4.0 and https://github.com/zwave-js/zwave-js-server/pull/846 for details


zwave_js_server/model/controller/firmware.py
```diff
@@ -0,0 +1,48 @@
...
+
+class FirmwareUpdateStatus(IntEnum):
+    """Enum with all controller firmware update status values.
+
+    https://zwave-js.github.io/node-zwave-js/#/api/node?id=status
+    """
+
+    ERROR_TIMEOUT = 0
+    # The maximum number of retry attempts for a firmware fragments were reached
+    ERROR_RETRYLIMITREACHED = 1
```
MartinHjelmare, line 14:
```suggestion
    ERROR_RETRY_LIMIT_REACHED = 1
```

zwave_js_server/model/controller/firmware.py
```diff
@@ -0,0 +1,48 @@
+"""Provide a model for Z-Wave controller firmware."""
+from enum import IntEnum
+from typing import TypedDict
+
+
+class FirmwareUpdateStatus(IntEnum):
```
MartinHjelmare, line 6: I think it's confusing to have two classes with the same name. I suggest we name the controller and node firmware update status differently.

raman325: I went back and forth on this. They are scoped to their import and represent the same thing in different use cases. We can add the extra info but in cases where they conflict the consumer should do that IMO

MartinHjelmare: I don't think it's a good idea. They don't have the same interface so they are not the same.

Same goes for the firmware update progress.

MartinHjelmare (approved):
Thanks!

I'm not sure about our documentation practice but I don't think we added the new firmware otw command to the readme for schema 25 in the server.

raman325:
> Thanks!
> 
> I'm not sure about our documentation practice but I don't think we added the new firmware otw command to the readme for schema 25 in the server.

yeah I missed docs updates there. Will update, thanks for the reminder 👍🏾

PR #568, opened 2023-02-03
Bump lib to 0.45.2

raman325:
This should be the last change needed for all the latest changes: https://github.com/home-assistant-libs/zwave-js-server-python/pull/567


raman325:
reopened because this is a feature that is part of the current schema so we should have a release that fixes this

MartinHjelmare (approved):

PR #569, opened 2023-02-03
Drop support for python 3.9

raman325:
Now that HA is minimum python 3.10, we can be too. I believe this means we can also use new style type annotations?


MartinHjelmare (approved):
Yes, most of the modern style type annotations are natively available in Python 3.10.

We should probably add Python 3.11 to CI as a follow up.

PR #571, opened 2023-02-05
Add python 3.11 support

raman325:
let me know if I missed anything but I think I added it everywhere I need to


MartinHjelmare (approved):
Looks good!

PR #581, opened 2023-02-18
Re-add optional firmware target to node.update_firmware command

raman325:
### Breaking Change
This PR removes the `FirmwareUpdateData` and `FirmwareUpdateDataDataType` classes which were used for both controllers and nodes and creates separate corresponding classes for each model type due to the differences in schemas.

### Description
From a conversation in #devs_zwave I learned that I shouldn't have removed the target option in https://github.com/zwave-js/zwave-js-server/pull/765 and https://github.com/home-assistant-libs/zwave-js-server-python/pull/503 so I am re-adding it

Dependent on https://github.com/zwave-js/zwave-js-server/pull/868 and a corresponding server bump which is upcoming for this next release cycle


zwave_js_server/model/controller/firmware.py
```diff
@@ -1,6 +1,36 @@
...
-from typing import TypedDict
+from typing import Optional, TypedDict
+
+from ...util.helpers import convert_bytes_to_base64
+
+
+class ControllerFirmwareUpdateDataDataType(TypedDict, total=False):
+    """Represent a controller firmware update data dict type."""
+
+    filename: str  # required
```
MartinHjelmare, line 12: We could import `Required` from `typing_extensions` and use here. It's available by default in Python 3.11.

```py
from typing_extensions import Required
```

https://peps.python.org/pep-0655/

MartinHjelmare (approved):

zwave_js_server/model/controller/firmware.py
raman325: should I do that here or do that across the board in another PR? I was reviewing the PEPs this weekend and I want to switch to union typing and generic typing in addition to more strongly typing our typeddicts. It's a lot so would have to be broken up over several PRs

MartinHjelmare: We can do it in one or more follow ups. :+1:

PR #606, opened 2023-03-19
PEP 585: Use generic typing

### LINKED ISSUE #589: PEP 585: Use generic typing
e.g. `dict` instead of `Dict`

https://peps.python.org/pep-0585/

raman325:
to implement, I went through the following types, removed the import, and then did a find/replace on the whole word:
- List
- DefaultDict
- Dict
- Tuple

fixes #589


MartinHjelmare (approved):
Nice!

Issue #588, opened 2023-02-20
PEP 604: Use union typing


raman325:
e.g. `int | None` instead of `Optional[int]`

https://peps.python.org/pep-0604/

PR #609, opened 2023-03-21
PEP 604: Use union typing

raman325:
Fixes https://github.com/home-assistant-libs/zwave-js-server-python/issues/588

Note that I had to disable the `line-too-long` pylint directive because of the changes here: https://github.com/home-assistant-libs/zwave-js-server-python/compare/master...raman325:zwave-js-server-python:pep604?expand=1#diff-010ebdc27781788869d8d79d7ea606ef0b2c7b9faaf59384af48675e0f614d45R108


zwave_js_server/model/node/event_model.py
```diff
@@ -103,11 +105,7 @@ class NotificationEventModel(BaseNodeEventModel):
 
     event: Literal["notification"]
     ccId: CommandClass
-    args: Union[
-        NotificationNotificationArgsDataType,
-        EntryControlNotificationArgsDataType,
-        PowerLevelNotificationArgsDataType,
-    ]
+    args: NotificationNotificationArgsDataType | EntryControlNotificationArgsDataType | PowerLevelNotificationArgsDataType
```
MartinHjelmare, line 108: I think we can use parenthesis to avoid a too long line.

MartinHjelmare (approved):

zwave_js_server/model/node/event_model.py
raman325: thanks, weird that black didn't do that automatically

PR #624, opened 2023-03-29
Remove assertions that will raise unhandled exceptions

raman325:
@kpine discovered an issue where when the Client first connects, it attempts to create the Driver, which in turn creates the Controller, which in turn creates all the Nodes, which in turn creates `NodeStatistics` for each node. `NodeStatistics` gets to the controller through `client.driver` in order to look up nodes for the `<lwr/nlwr>.route_failed_between` and `<lwr/nlwr>.repeaters` properties, and we assert `client.driver` as part of that look up because it's an optional attribute. Because all of this is part of the Driver creation, the Driver hasn't been set on the Client yet and the assertion fails. This causes a disconnect which then loops to a reconnect and the same issue occurs. This will only occur if the network has already been up and at least one node has a last working router or next to last working route stored.

While investigating this I found another place where the assertion seemed unnecessary, so removed it there too.

### DIFF
--- a/zwave_js_server/model/statistics.py
+++ b/zwave_js_server/model/statistics.py
@@ -55,7 +55,7 @@ class RouteStatistics:
             for node_id in self.data["repeaters"]
         ]
 
-    @property
+    @cached_property
     def rssi(self) -> int | None:
         """Return RSSI."""
         if (rssi := self.data.get("rssi")) is None:
@@ -64,7 +64,7 @@ class RouteStatistics:
             raise RssiErrorReceived(RssiError(rssi))
         return rssi
 
-    @property
+    @cached_property
     def repeater_rssi(self) -> list[int]:
         """Return repeater RSSI."""
         repeater_rssi = self.data.get("repeaterRSSI", [])
--- a/zwave_js_server/model/statistics.py
+++ b/zwave_js_server/model/statistics.py
@@ -2,6 +2,7 @@
 from __future__ import annotations
 
 from dataclasses import dataclass, field
+from functools import cached_property
 from typing import TYPE_CHECKING, TypedDict
 
 from zwave_js_server.exceptions import RepeaterRssiErrorReceived, RssiErrorReceived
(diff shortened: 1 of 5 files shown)


raman325:
I'm wondering if the right thing to do here is to pass the controller directly to `RouteStatistics` to avoid this approach. That would require a pretty significant refactor because the Node would have to store the controller as well, so maybe a TODO for later if you agree that's a better approach.

MartinHjelmare:
I'm missing a description of the problem that the PR is solving.

raman325:
whoops sorry, talking about it in Discord and forgot to redescribe it here. Added

MartinHjelmare:
I don't understand the question fully. All the nodes haven't been instantiated when one node is instantiated as that happens in a for loop. We can't look up other nodes until we know that the driver is ready as that's the point when everything has been instantiated from the received driver state.

What is your suggestion?

MartinHjelmare (approved):
This PR looks good.

raman325:
> I don't understand the question fully. All the nodes haven't been instantiated when one node is instantiated as that happens in a for loop. We can't look up other nodes until we know that the driver is ready as that's the point when everything has been instantiated from the received driver state.
> 
> 
> 
> What is your suggestion?

Fair point, maybe we just need to move these back to property functions so they're accessed on demand. That would bypass the issue

zwave_js_server/model/statistics.py
```diff
@@ -40,13 +40,14 @@ class RouteStatistics:
...
-    repeaters: list["Node"] = field(init=False)
+    repeaters: list["Node"] = field(init=False, default_factory=list)
     route_failed_between: tuple["Node", "Node"] | None = field(init=False, default=None)
 
     def __post_init__(self) -> None:
         """Post initialize."""
         self.protocol_data_rate = ProtocolDataRate(self.data["protocolDataRate"])
-        assert self.client.driver
+        if not self.client.driver:
+            return
```
balloob, line 50: This is not good, because it means the object will contain incorrect data. 

Your idea of a function is better, as it is calculated. You can also do `@property` and then you don't need code changes.

You can also turn it into a `@cached_property` if you know it's never going to change, that way it's only calculated once during lifetime of the object.

PR #626, opened 2023-03-31
Fix bug in CommandClassInfo model

raman325:
from https://github.com/home-assistant/core/issues/90551 I noticed that we are converting id to a `CommandClass` in the property which doesn't make sense so added a separate property to do it


MartinHjelmare:
Do we need any tests?

raman325:
done

firstof9:
Looks like black needs to run against the updated test

Issue #607, opened 2023-03-19
Validation for `async_multicast_set_value` is too strict


kelchm:
The value validation implemented in `async_multicast_set_value` ([here](https://github.com/home-assistant-libs/zwave-js-server-python/blob/5f0a20dfebc1977cec29a333c852af380d82eb18/zwave_js_server/util/multicast.py#L50-L51)) seems to be a bit too strict and results in the rejection of commands that are supported by the underlying zwave-js API. 

This causes usability issues with devices which do _not_ have their root endpoint exposed by default (such as the [Fibaro FGD212](https://gist.github.com/kelchm/0f37b39d486d94061f5ac81a059c809d)). For example, when targeting an area which contains a device like this one with a multicast command, an error will be returned that the node does not have that value ([example](https://gist.github.com/kelchm/b0052d2f09df676db32fec79f7edf6bc)).

My understanding is that the capabilities of the root endpoint, at minimum, should always mirror mirror endpoint 1. Given this, would it be appropriate to either relax this validation to allow for it for it to be explicitly skipped?

PR #643, opened 2023-04-11
Remove validation on multicast set value command

raman325:
In reviewing https://github.com/home-assistant-libs/zwave-js-server-python/issues/607 with @kpine and @AlCalzone we came to the conclusion that we should let the driver do any validation that is necessary.

This fixes https://github.com/home-assistant-libs/zwave-js-server-python/issues/607

### DIFF
--- a/test/util/test_multicast.py
+++ b/test/util/test_multicast.py
@@ -1,8 +1,5 @@
 """Test node utility functions."""
-import pytest
-
 from zwave_js_server.const import CommandClass
-from zwave_js_server.exceptions import NotFoundError
 from zwave_js_server.util.multicast import (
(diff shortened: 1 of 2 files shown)


MartinHjelmare (commented):
What will happen now if incorrect parameters are sent to the driver?

raman325:
> What will happen now if incorrect parameters are sent to the driver?

the driver handles the validation entirely with this change

MartinHjelmare:
There's no `ZwaveError` in this library. We'll raise a `FailedZWaveCommand` if we get a `zwave_error` code in the result.

We don't currently handle this in the service handler in the core integration. We should improve this in the future.

MartinHjelmare (approved):

PR #651, opened 2023-04-29
Add support for bitfield config parameters

raman325:
WIP, need more details


raman325:
this is not needed because bitfield parameters are now exposed as partial parameters

PR #663, opened 2023-05-19
Fix typehint for Value Metadata states

raman325:
The below issues are happening because of this PR: https://github.com/zwave-js/node-zwave-js/pull/5792

Even though the original state keys are numeric, they are represented as strings so we should fix that typehint as well

Fixes https://github.com/home-assistant/core/issues/93252
Fixes https://github.com/home-assistant/core/issues/93299

### DIFF
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -81,14 +81,6 @@ def invalid_multilevel_sensor_type_state_fixture():
     return json.loads(load_fixture("invalid_multilevel_sensor_type_state.json"))
 
 
-@pytest.fixture(name="client_session")
-def client_session_fixture(ws_client):
-    """Mock an aiohttp client session."""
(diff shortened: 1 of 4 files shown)


MartinHjelmare:
Is this backwards compatible?

raman325:
It should be as our tests pass. Just to be sure, I changed the typehint from `dict[str | bool, str]` to `dict[None, str]` and three tests failed. If we want to be safe we can add `int` as a typehint but the state key has always been a string representation of the int as far as I have seen

MartinHjelmare:
Should we add a test with the metadata in the reported issues?

raman325:
> Should we add a test with the metadata in the reported issues?

done

MartinHjelmare (approved):

PR #665, opened 2023-05-19
Handle device class data is undefined

raman325:
Fixes https://github.com/home-assistant/core/issues/93096


zwave_js_server/model/node/__init__.py
```diff
@@ -138,7 +138,7 @@ def index(self) -> int:
     @property
     def device_class(self) -> DeviceClass:
```
MartinHjelmare, line 139: Maybe this should return None instead and we don't change the properties of the `DeviceClass` class?

MartinHjelmare (approved):

PR #669, opened 2023-05-24
Bump zwave-js-server-python to 0.48.2

raman325:
Because endpoint was added to the end of the function in https://github.com/home-assistant-libs/zwave-js-server-python/pull/668, the change is backwards compatible so this can be a patch release

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -5,7 +5,7 @@ from setuptools import find_packages, setup
 
 PROJECT_DIR = Path(__file__).parent.resolve()
 README_FILE = PROJECT_DIR / "README.md"
-VERSION = "0.48.1"
+VERSION = "0.48.2"
 
 
 setup(


MartinHjelmare:
We don't think non 0 endpoint config parameter support is a new feature but a fix?

raman325:
@AlCalzone is "config parameters not on endpoint 0" a new feature, and if so, which release would it be added with?

raman325:
> We don't think non 0 endpoint config parameter support is a new feature but a fix?

If it's not a bugfix, I would consider it a code quality improvement rather than a new feature

AlCalzone:
I'm treating it as a breaking change (although you might consider it a feature), so it will land in v11.

raman325:
> It's not code quality as we didn't support it before. Or am I missing something?

PR #676, opened 2023-05-30
Add Energy Production CC constants and util functions

raman325:
Support for the Energy Production CC which was introduced in zwave-js v10.

### DIFF
--- a/zwave_js_server/util/command_class/energy_production.py
+++ b/zwave_js_server/util/command_class/energy_production.py
@@ -1,0 +1,39 @@
+"""Energy Production Command Class specific utility functions for values."""
+from __future__ import annotations
+
+from ...const import CommandClass
+from ...const.command_class.energy_production import (
+    CC_SPECIFIC_PARAMETER,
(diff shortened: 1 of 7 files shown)


zwave_js_server/util/command_class/energy_production.py
```diff
@@ -0,0 +1,39 @@
...
+def get_energy_production_parameter(value: Value) -> EnergyProductionParameter:
+    """Get the EnergyProductionParameter for a given value."""
+    if value.command_class != CommandClass.ENERGY_PRODUCTION:
+        raise InvalidCommandClass(value, CommandClass.ENERGY_PRODUCTION)
+    try:
+        return EnergyProductionParameter(
+            value.metadata.cc_specific[CC_SPECIFIC_PARAMETER]
+        )
+    except ValueError:
+        raise UnknownValueData(  # pylint: disable=raise-missing-from
```
MartinHjelmare, line 25: I think you can use `raise from None` if you want to disable the exception chaining.

MartinHjelmare (approved):

PR #677, opened 2023-05-30
Bump zwave-js-server-python to 0.49.0

raman325:
We need to release https://github.com/home-assistant-libs/zwave-js-server-python/pull/676 for a PR I am about to submit in HA core that I'd like to get in ahead of the beta, but per another conversation, we are bumping minor because of the change to support config parameters that are not on endpoint 0. We may have to do a patch bump in the future to raise the schema version so that we can push users onto v11.


MartinHjelmare (approved):
Raising the minimum schema version of the library should also be a minor release as it's a breaking change.

PR #680, opened 2023-06-11
Add support for zwave-js 11 (schema 29)

raman325:
See https://github.com/zwave-js/zwave-js-server/pull/956 for more details

In addition to what's listed there, we added support for the boolean configuration value type. This is a breaking change as it introduces a new `ConfigurationValueType` boolean which changes the `ConfigurationValueType` for existing cases.

### DIFF
--- a/zwave_js_server/model/controller/__init__.py
+++ b/zwave_js_server/model/controller/__init__.py
@@ -375,7 +375,7 @@ class Controller(EventBase):
         self, strategy: ExclusionStrategy | None = None
     ) -> bool:
         """Send beginExclusion command to Controller."""
-        payload: dict[str, str | ExclusionStrategy] = {
+        payload: dict[str, str | dict[str, ExclusionStrategy]] = {
             "command": "controller.begin_exclusion"
         }
         if strategy is not None:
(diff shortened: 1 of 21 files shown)


zwave_js_server/model/duration.py
```diff
@@ -0,0 +1,33 @@
...
+    unit: Literal["seconds", "minutes", "unknown", "default"] = field(init=False)
+    value: int | float | None = field(init=False)
+
+    def __post_init__(self) -> None:
+        """Post init."""
+        self.unit = self.data["unit"]
+        self.value = self.data.get("value")
+
+    def __repr__(self) -> str:
+        """Return the representationt."""
```
MartinHjelmare, line 30:
```suggestion
        """Return the representation."""
```

zwave_js_server/model/controller/__init__.py
```diff
@@ -793,7 +794,7 @@ async def async_firmware_update_ota(
             },
             require_schema=24,
```
OXINARF, line 795: Doesn't this now require schema 29? Otherwise it can't parse the return data (see my comment on server PR too).

raman325: the require_schema attribute is effectively a no-op now that the minimum schema version matches the maximum schema version, but you are right

OXINARF: True, I know, but I just thought that if it exists, then we should use it properly.

OXINARF (commented):
A couple of places missing from the require_schema

MartinHjelmare (approved):

PR #702, opened 2023-07-17
Make Node model more performant

raman325:
Instead of keeping all `Value` and `Endpoint` data in the `Node` data, we will only store it in each `Value` and `Endpoint` so there is only one place to keep it up to date. Callers who want to get a full state dump of a Node will need to now combine the data from these three models (or use the dump feature)

### DIFF
--- a/zwave_js_server/model/node/__init__.py
+++ b/zwave_js_server/model/node/__init__.py
@@ -1,6 +1,7 @@
 """Provide a model for the Z-Wave JS node."""
 from __future__ import annotations
 
+import copy
 import logging
 from datetime import datetime
 from typing import TYPE_CHECKING, Any, cast
@@ -100,7 +101,7 @@ class Node(EventBase):
         """Initialize the node."""
         super().__init__()
         self.client = client
-        self.data: NodeDataType = {}
+        self._data: NodeDataType = {}
         self._device_config = DeviceConfig({})
         self._statistics = NodeStatistics(
             client, data.get("statistics", DEFAULT_NODE_STATISTICS)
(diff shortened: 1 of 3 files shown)


zwave_js_server/model/node/__init__.py
```diff
@@ -332,70 +344,75 @@ def firmware_update_progress(self) -> NodeFirmwareUpdateProgress | None:
...
-        value_ids = (_get_value_id_str_from_dict(self, val) for val in data["values"])
-        self.values = {
-            value_id: val
-            for value_id, val in self.values.items()
-            if value_id in value_ids
+        new_values_data = {
+            _get_value_id_str_from_dict(self, val): val
+            for val in self._data.pop("values")
         }
+        stale_value_ids = list(set(self.values) - set(new_values_data))
```
MartinHjelmare, line 373: We don't need to create lists.
```suggestion
        stale_value_ids = set(self.values) - set(new_values_data)
```

zwave_js_server/model/node/__init__.py
```diff
@@ -332,70 +344,75 @@ def firmware_update_progress(self) -> NodeFirmwareUpdateProgress | None:
...
+        remaining_value_ids = list(set(new_values_data) - set(stale_value_ids))
```
MartinHjelmare, line 374:
```suggestion
        remaining_value_ids = set(new_values_data) - set(stale_value_ids)
```

zwave_js_server/model/node/__init__.py
```diff
@@ -332,70 +344,75 @@ def firmware_update_progress(self) -> NodeFirmwareUpdateProgress | None:
...
 
-        # Remove stale endpoints
-        self.endpoints = {
-            idx: endpoint
-            for idx, endpoint in self.endpoints.items()
-            if idx in (endpoint["index"] for endpoint in self.data["endpoints"])
+        new_endpoints_data = {
+            endpoint["index"]: endpoint for endpoint in self._data.pop("endpoints")
         }
+        stale_endpoint_idxs = list(set(self.endpoints) - set(new_endpoints_data))
```
MartinHjelmare, line 395:
```suggestion
        stale_endpoint_idxs = set(self.endpoints) - set(new_endpoints_data)
```

zwave_js_server/model/node/__init__.py
```diff
@@ -901,24 +909,14 @@ def handle_value_updated(self, event: Event) -> None:
...
             value = _init_value(self, evt_val_data)
             self.values[value.value_id] = event.data["value"] = value
-            self.data["values"].append(evt_val_data)
         else:
             value.receive_event(event)
             event.data["value"] = value
-            self.data["values"][self.value_data_idx(value_id)].update(evt_val_data)
-
-        node_val_data = self.data["values"][self.value_data_idx(value_id)]
-        if "newValue" in evt_val_data:
```
MartinHjelmare, line 911: Is this `newValue` item handled in the property `data` that we now expose?

zwave_js_server/model/node/__init__.py
```diff
@@ -126,134 +127,145 @@ def __eq__(self, other: object) -> bool:
             self.client.driver == other.client.driver and self.node_id == other.node_id
         )
 
+    @property
+    def data(self) -> NodeDataType:
```
MartinHjelmare, line 131: What currently needs this that we know? If we're making a breaking change anyway, maybe we don't need this?

MartinHjelmare: Yes, sounds good.

zwave_js_server/model/node/__init__.py
raman325: no longer relevant since we removed the new property

zwave_js_server/model/node/__init__.py
```diff
@@ -348,54 +349,57 @@ def keep_awake(self) -> bool:
...
-            for value_id, val in self.values.items()
-            if value_id in value_ids
+        new_values_data = {
+            _get_value_id_str_from_dict(self, val): val
+            for val in self.data.pop("values")
         }
+        new_value_ids = set(new_values_data)
+
+        # Remove stale values
+        for value_id in (stale_value_ids := set(self.values) - new_value_ids):
```
MartinHjelmare, line 365: I don't see the value in a walrus here. We'll always use the local variable.

raman325:
> The PR is still breaking since we don't update the `data`attribute anymore, right?

Yeah that's fair. We pop the values out so there is no stale data but that is still a breaking change

zwave_js_server/model/node/__init__.py
raman325: Walrus is because it's referenced in the next code block to get the remaining ids to update/create. It may be more clear that's the case with the last commit rather than setting it using a walrus

MartinHjelmare: Right, but we don't need to use a walrus. Use a separate line above.

zwave_js_server/model/value.py
```diff
@@ -266,19 +266,18 @@ def receive_event(self, event: Event) -> None:
...
     def update(self, data: ValueDataType) -> None:
         """Update data."""
+        data.pop("prevValue", None)
         self.data.update(data)
-        if "newValue" in data:
-            self._value = self.data["value"] = data["newValue"]
-        elif "value" in data:
-            self._value = self.data["value"] = data["value"]
+        if (new_value := self.data.pop("newValue", None)) is not None:
+            self._value = self.data["value"] = new_value
```
MartinHjelmare, line 272:
```suggestion
            self.data["value"] = new_value
```

MartinHjelmare (approved):
Looks good! Just update the PR description.

PR #703, opened 2023-07-17
Bump pydantic to 2.0 with backwards compatibility

raman325:
Will allow HA to update to pydantic 2.0 while we wait for it to be default so we can leverage 2.0 features


frenck (approved):
LGTM, Thanks @raman325 👍 

../Frenck

PR #707, opened 2023-07-23
Drop Python 3.10 support

raman325:
HA has dropped Python 3.10 support so we should as well, it will also avoid us having to include a backport in https://github.com/home-assistant-libs/zwave-js-server-python/pull/705


setup.py
```diff
@@ -33,7 +33,6 @@
         "Natural Language :: English",
         "Programming Language :: Python :: 3",
         "Programming Language :: Python :: 3.10",
```
MartinHjelmare, line 35: Wrong line was removed.

MartinHjelmare (approved):
Thanks!

PR #713, opened 2023-07-29
Fix bugs from #702

raman325:
Two changes:
1. To preserve ordering, `Node.update` should add values in the same order as the server on load
2. Fix issue with `Value.update` - when popping `prevValue`, we were removing it from the event data dictionary. We should preserve the event data as much as possible and instead remove it from a Value's data attribute

### DIFF
--- a/zwave_js_server/model/node/__init__.py
+++ b/zwave_js_server/model/node/__init__.py
@@ -366,8 +366,14 @@ class Node(EventBase):
         for value_id in stale_value_ids:
             self.values.pop(value_id)
 
-        # Updating existing values and populate new values
-        for value_id in new_value_ids - stale_value_ids:
+        # Updating existing values and populate new values. Preserve value order if
+        # initializing values for the node for the first time by using the key order
+        # which is deterministic
+        for value_id in (
+            new_value_ids - stale_value_ids
+            if stale_value_ids
+            else list(new_values_data)
+        ):
             val = new_values_data[value_id]
             try:
                 if value_id in self.values:
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,2 @@
-aiohttp==3.8.4
-pydantic==2.0.3
+aiohttp==3.8.5
+pydantic>=1.10.0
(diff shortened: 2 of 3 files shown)


zwave_js_server/model/node/__init__.py
```diff
@@ -366,8 +366,14 @@ def update(self, data: NodeDataType) -> None:
         for value_id in stale_value_ids:
             self.values.pop(value_id)
 
-        # Updating existing values and populate new values
-        for value_id in new_value_ids - stale_value_ids:
+        # Updating existing values and populate new values. Preserve value order if
```
MartinHjelmare, line 369: Why does the order matter?

raman325: Not sure yet but without this change existing HA tests were failing presumably because it impacted the way entities got named and the same state calls from before were getting different values.

I opened a TODO issue so I don't lose track of this, but for now this will avoid a breaking change. The proper fix may end up being a breaking change on the HA side

MartinHjelmare (approved):
I'm a bit scared for the whole change now as there seems to be unknown impact.

raman325:
We could revert the original PR and reimplement this in a new one with this in mind. But this is definitely an improvement over what's in master if we opt not to revert

MartinHjelmare:
I think we need to find out why the order matters to know how to proceed.

raman325:
I believe this is why it matters: https://github.com/home-assistant/core/blob/dev/homeassistant/components/zwave_js/entity.py#L187-L192

I'm not sure why the endpoint has to be greater than 1 meaning it excludes 1. Was that supposed to be greater than 0? This means that the same value on endpoints 0 and 1 will have the same name. One of them will have to have a 1 in the entity ID to avoid collisions but it will depend on the order in which the two values are discovered?

raman325:
in conclusion I think the right thing to do is to update that conditional in the core naming logic to be > 0 instead of > 1. It may cause some name changes but it's less brittle

MartinHjelmare:
Ok, sounds good.

PR #719, opened 2023-08-09
Improve logic for determining config value boolean type

raman325:
I realized from https://github.com/home-assistant/core/pull/94760#discussion_r1287923508 that our check here isn't correct as values that are boolean do not have mins and maxes.

### DIFF
--- a/zwave_js_server/model/value.py
+++ b/zwave_js_server/model/value.py
@@ -2,6 +2,7 @@
 from __future__ import annotations
 
 from dataclasses import dataclass, field
+from enum import StrEnum
 from typing import TYPE_CHECKING, Any, TypedDict
 
 from ..const import VALUE_UNKNOWN, CommandClass, ConfigurationValueType, SetValueStatus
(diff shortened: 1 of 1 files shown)


zwave_js_server/model/value.py
```diff
@@ -301,8 +311,9 @@ def configuration_value_type(self) -> ConfigurationValueType:
         max_ = self.metadata.max
         states = self.metadata.states
         allow_manual_entry = self.metadata.allow_manual_entry
+        type_ = self.metadata.type
 
-        if max_ == 1 and min_ == 0 and not states:
+        if ((max_ == 1 and min_ == 0) or type_ == ValueType.BOOLEAN) and not states:
```
MartinHjelmare, line 316:
```suggestion
        if (max_ == 1 and min_ == 0 or type_ == ValueType.BOOLEAN) and not states:
```

MartinHjelmare (approved):

PR #729, opened 2023-08-22
Cancel waiting for a response when node goes to dead/asleep

raman325:
We should not wait for a response when we send a command and the node that we are sending a command for moves to a status that means we won't get a response in a reasonable time

### DIFF
--- a/zwave_js_server/model/node/__init__.py
+++ b/zwave_js_server/model/node/__init__.py
@@ -466,7 +466,8 @@ class Node(EventBase):
             kwargs["require_schema"] = require_schema
 
         if wait_for_result or (
-            wait_for_result is None and self.status != NodeStatus.ASLEEP
+            wait_for_result is None
+            and self.status not in (NodeStatus.ASLEEP, NodeStatus.DEAD)
         ):
             result_task = asyncio.create_task(
                 self.client.async_send_command(message, **kwargs)
@@ -907,7 +908,7 @@ class Node(EventBase):
     def handle_wake_up(self, event: Event) -> None:
         """Process a node wake up event."""
         # pylint: disable=unused-argument
-        self._status_event.release()
+        self._status_event.clear()
         self.data["status"] = NodeStatus.AWAKE
 
     def handle_sleep(self, event: Event) -> None:
@@ -925,7 +926,7 @@ class Node(EventBase):
     def handle_alive(self, event: Event) -> None:
         """Process a node alive event."""
         # pylint: disable=unused-argument
-        self._status_event.release()
+        self._status_event.clear()
         self.data["status"] = NodeStatus.ALIVE
 
     def handle_interview_started(self, event: Event) -> None:
--- a/zwave_js_server/model/node/__init__.py
+++ b/zwave_js_server/model/node/__init__.py
@@ -478,7 +478,7 @@ class Node(EventBase):
             if not self._status_lock.locked() and not result_task.done():
                 result_task.cancel()
                 return None
-            return await result_task.result()
+            return result_task.result()
 
         await self.client.async_send_command_no_wait(message, **kwargs)
         return None
--- a/zwave_js_server/model/node/__init__.py
+++ b/zwave_js_server/model/node/__init__.py
@@ -475,9 +475,7 @@ class Node(EventBase):
             result_task = asyncio.create_task(
                 self.client.async_send_command(message, **kwargs)
             )
-            status_task = asyncio.create_task(
-                self._status_event.wait()
-            )
+            status_task = asyncio.create_task(self._status_event.wait())
             await asyncio.wait(
                 [result_task, status_task],
                 return_when=asyncio.FIRST_COMPLETED,
--- a/zwave_js_server/model/node/__init__.py
+++ b/zwave_js_server/model/node/__init__.py
@@ -475,11 +475,13 @@ class Node(EventBase):
                 [result_task, self._status_event.wait()],
                 return_when=asyncio.FIRST_COMPLETED,
             )
-            if self._status_event.is_set() and not result_task.done():
+            try:
+                if self._status_event.is_set() and not result_task.done():
+                    result_task.cancel()
+                    return None
+                return result_task.result()
+            finally:
                 self._status_event.clear()
-                result_task.cancel()
-                return None
-            return result_task.result()
 
         await self.client.async_send_command_no_wait(message, **kwargs)
         return None
--- a/zwave_js_server/model/node/__init__.py
+++ b/zwave_js_server/model/node/__init__.py
@@ -1,6 +1,7 @@
 """Provide a model for the Z-Wave JS node."""
 from __future__ import annotations
 
+import asyncio
 import copy
 import logging
 from datetime import datetime
@@ -113,6 +114,7 @@ class Node(EventBase):
         self._firmware_update_progress: NodeFirmwareUpdateProgress | None = None
         self.values: dict[str, ConfigurationValue | Value] = {}
         self.endpoints: dict[int, Endpoint] = {}
+        self._status_lock = asyncio.Lock()
         self.update(data)
 
     def __repr__(self) -> str:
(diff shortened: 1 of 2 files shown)


zwave_js_server/model/node/__init__.py
```diff
@@ -893,21 +907,25 @@ def handle_check_route_health_progress(self, event: Event) -> None:
     def handle_wake_up(self, event: Event) -> None:
         """Process a node wake up event."""
         # pylint: disable=unused-argument
+        self._status_event.release()
```
MartinHjelmare, line 910:
```suggestion
        self._status_event.clear()
```

zwave_js_server/model/node/__init__.py
```diff
@@ -466,8 +468,20 @@ async def async_send_command(
         if wait_for_result or (
             wait_for_result is None and self.status != NodeStatus.ASLEEP
```
MartinHjelmare, line 469: Why aren't we checking for dead status too here?

raman325: good question? no idea...

zwave_js_server/model/node/__init__.py
```diff
@@ -466,8 +468,20 @@ async def async_send_command(
...
+                [result_task, self._status_event.wait()],
+                return_when=asyncio.FIRST_COMPLETED,
+            )
+            try:
+                if self._status_event.is_set() and not result_task.done():
+                    result_task.cancel()
+                    return None
+                return result_task.result()
+            finally:
+                self._status_event.clear()
```
MartinHjelmare, line 485: What's the reason for clearing here after the call? Is it to allow calls with `wait_for_result` set to True, after a node has already changed status to eg dead?

zwave_js_server/model/node/__init__.py
```diff
@@ -464,10 +466,23 @@ async def async_send_command(
...
-            wait_for_result is None and self.status != NodeStatus.ASLEEP
+            wait_for_result is None
+            and self.status not in (NodeStatus.ASLEEP, NodeStatus.DEAD)
         ):
-            result = await self.client.async_send_command(message, **kwargs)
-            return result
+            result_task = asyncio.create_task(
+                self.client.async_send_command(message, **kwargs)
+            )
+            await asyncio.wait(
```
MartinHjelmare, line 475: Should we ignore the event if `wait_for_result` is set to True?

zwave_js_server/model/node/__init__.py
raman325: just seems like good hygiene? either we got here because the status event was set or we got here because the response was returned. We clear the event on success so that the net command awaits for the next status change or a response

zwave_js_server/model/node/__init__.py
raman325: no I think that's one of the core issues. We can't afford to wait for a response when a device is asleep/dead, otherwise it will block forever. So `wait_for_result` shouldn't matter for most commands that are sent

zwave_js_server/model/node/__init__.py
MartinHjelmare: Yeah, but if the event was set no new calls will get here unless they are made with `wait_for_result` set to True, until the event is cleared by the node status.

raman325: so do you consider this a no-op then?

zwave_js_server/model/node/__init__.py
MartinHjelmare: Then the usefulness of this parameter isn't clear to me. Why should we be able to start waiting for a command to a node that is sleeping or is dead if we won't finish waiting for the command if the node fell asleep or died?

raman325: yeah that's a fair point and I'm not sure how to address it. There are commands that need a response, but there are also commands sent when the node status is effectively incorrect. So what do we do when it corrects itself? maybe we should only cancel if `wait_for_result` is set to `None`?

zwave_js_server/model/node/__init__.py
MartinHjelmare: Not at the moment since we have `wait_for_result`.

raman325: OK I think I get it now. We want the event to be set anytime the node is in the ASLEEP or DEAD status and cleared anytime the node is in the AWAKE or ALIVE status , and we don't want to mess with the event in the command handler  👍🏾

zwave_js_server/model/node/__init__.py
raman325: OK I guess we should not follow this path when `wait_for_result` is `True`. Worst case scenario, nothing changes for certain calls. I looked through the calls that use it and I don't see anything that would block your typical services, those are almost always going to be set_value calls (the one exception I can find is for retrieving firmware information but I doubt anyone is using that service in an automation, or if they are, they probably shouldn't)

zwave_js_server/model/node/__init__.py
```diff
@@ -463,11 +465,28 @@ async def async_send_command(
...
+                self.client.async_send_command(message, **kwargs)
+            )
+            status_task = asyncio.create_task(
+                self._status_event.wait()
+            )
+            await asyncio.wait(
+                [result_task, status_task],
+                return_when=asyncio.FIRST_COMPLETED,
+            )
+            status_task.cancel()
```
raman325, line 485: OK so `asyncio.wait` wants Tasks, not coroutine. So now I wrapped the status event wait coro in a Task. It's appropriate to cancel the task here right?

zwave_js_server/model/node/__init__.py
```diff
@@ -463,11 +465,26 @@ async def async_send_command(
...
+            )
+            status_task = asyncio.create_task(self._status_event.wait())
+            await asyncio.wait(
+                [result_task, status_task],
+                return_when=asyncio.FIRST_COMPLETED,
+            )
+            status_task.cancel()
+            if self._status_event.is_set() and not result_task.done():
+                result_task.cancel()
+                return None
```
MartinHjelmare, line 486: Shouldn't we raise an error if the node died or fell asleep unexpectedly when waiting for a result?

zwave_js_server/model/node/__init__.py
MartinHjelmare: I think so.

zwave_js_server/model/node/__init__.py
raman325: @AlCalzone what do you think? If at the point where we sent a command to the driver, the node status was awake or alive, but then while waiting for a response, the node status switched to asleep or dead, should we consider that unexpected behavior? Or should we expect that the driver will queue the command and will eventually issue it? I would think that if an error needed to be raised, it would be the Driver's responsibility to do so, no? The driver currently isn't doing that, so that may be a clue as to what the right answer is, but it's worth explicitly asking.

Likewise, if we must wait for a response from a command because the response is the primary purpose of issuing the command (e.g. checking for firmware update capabilities), what should happen when this scenario occurs? Should we continue waiting and hope the node comes back to an alive or awake state, or should we raise an error?

AlCalzone: The driver should not silently drop commands that expect a response. If it does, that's a driver bug - might have to do with the stalling firmware updates that have been reported recently.

> but then while waiting for a response, the node status switched to asleep or dead, should we consider that unexpected behavior?

IMO, no. A node can be assumed alive or awake, but fail to respond the moment we send a command. Especially with sleeping nodes, this has to be expected. If this is a mains-powered node, the driver should throw after giving up. If this is a node that can sleep, the driver should re-queue the command and eventually handle it.

Side note: I'm adding a callback applications can use to get notified about command progress updates:
https://github.com/zwave-js/node-zwave-js/pull/6212

raman325: thanks so then I think the answer @MartinHjelmare is we don't want to raise, instead we should treat it like it's a sleeping/dead node and not wait for the result (except for in cases where we must wait for the result). If the node had been asleep or dead when the command was issued that would have been the behavior anyway right?

raman325:
We should cherry pick this so we can release a patch fix ahead of the minor release which bumps the schema. What's the best way to do that? Create a separate branch that removes the new features and release from that? Will we mess up release drafter in the process?

MartinHjelmare:
It's Home Assistant beta release on Wednesday. I think we can aim for that. Not sure there will be more patch releases anyway.

raman325:
> It's Home Assistant beta release on Wednesday. I think we can aim for that. Not sure there will be more patch releases anyway.

I was thinking more for the sake of external users, if there are any. But I am ok with that

zwave_js_server/model/node/__init__.py
MartinHjelmare: Ok, good.

MartinHjelmare (approved):

PR #731, opened 2023-08-28
Add support for changes in zwave-js 11.12.0

raman325:
We haven't released the server yet and with it we introduced a new event that I believe we would have to handle: https://github.com/zwave-js/zwave-js-server/pull/1015


MartinHjelmare (approved):
Just needs tests.

PR #743, opened 2023-09-04
Handle maybe not known responses

raman325:
We have to assume that the response could be undefined, in which case the key wouldn't be in the deserialized dict because typescript does not serialize undefined


zwave_js_server/model/node/__init__.py
```diff
@@ -689,15 +691,17 @@ async def async_has_security_class(self, security_class: SecurityClass) -> bool:
             wait_for_result=True,
         )
         assert data
```
MartinHjelmare, line 693: This will still raise if `data` is empty.

MartinHjelmare (changes requested):

MartinHjelmare:
There are also two keys in the node state that have MaybeNotKnown types in the server. Do we handle those correctly here in the client?

raman325:
> There are also two keys, isSecure and lastSeen, in the node state that have MaybeNotKnown types in the server. Do we handle those correctly here in the client?

fixed `isSecure`, `lastSeen` was already handled correctly

MartinHjelmare (commented):
Is the fixture `is_secure_unknown` incorrect? Maybe update it or remove it?

Can we add tests for the new uncovered cases?

MartinHjelmare (approved):
Change looks good! Just the comments on the tests remaining, which are not critical.

raman325:
> Is the fixture `is_secure_unknown` incorrect? Maybe update it or remove it?
> 
> Can we add tests for the new uncovered cases?

Updated fixture name - for the uncovered paths, I can fake it but I don't have any real dumps that would result in this case. Do you think it's worth manipulating existing state fixtures to achieve the same thing?

raman325:
let me know your thoughts, I will submit a second PR for tests if needed

MartinHjelmare:
I think it would be good to test it. We can adjust the node data before creating the node or similar.

Issue #738, opened 2023-08-29
Investigate wait_for_result behaviors


raman325:
I found three cases where we don't seem to act appropriately.
- async_abort_health_check - not checking result - we should probably return the result
- async_set_date_and_time - just returning success - do we need to wait for this?
- async_abort_firmware_update - not checking result - there's no result to wait for as documented

We seem to set wait_for_result when there's a return value we need. Considering this my thoughts are:
- We should probably raise an error immediately if the node is dead.
- What can we expect from sleeping nodes? Is it useful to wait for a result?
- We should probably raise an error if we're waiting for a result and the node dies or goes to sleep. The command is not likely to succeed at that point. At least not for dead nodes. I'm not sure what the driver strategy is for sleeping nodes. See my question above.

raman325:
> async_abort_health_check - not checking result - we should probably return the result

There's nothing to return so I would suggest we just don't wait for it

> async_set_date_and_time - just returning success - do we need to wait for this?

We could treat this like `setValue` and let the caller decide, otherwise default to whether the device is battery or not.

> async_abort_firmware_update - not checking result - there's no result to wait for as documented

no point waiting for result

> We seem to set wait_for_result when there's a return value we need. Considering this my thoughts are:
> 
> * We should probably raise an error immediately if the node is dead.

Why? As we've learned in other discussions, sometimes a node that is marked as dead is not dead. If we throw out commands because the node is marked as dead we will end up throwing out legitimate/receivable commands.

> * What can we expect from sleeping nodes? Is it useful to wait for a result?

There may be some cases where it is not useful to wait, especially when there are callbacks for the activity anyway, e.g. `testPowerLevel`. I don't think we can make a blanket statement on this, rather we need to handle it case by case

> * We should probably raise an error if we're waiting for a result and the node dies or goes to sleep. The command is not likely to succeed at that point. At least not for dead nodes. I'm not sure what the driver strategy is for sleeping nodes. See my question above.

Can we say that safely? @AlCalzone ?

CC @MartinHjelmare

MartinHjelmare:
I think point 4 and 6 were already answered in the previous PR.

AlCalzone:
> * We should probably raise an error immediately if the node is dead.

Please let the driver handle this. The node status in `node-zwave-js` is tracked on a best-effort basis, and `dead` is often just temporary. But with the additional checks in HA, getting out of this temporary state is unnecessarily complicated.

[...]

PR #744, opened 2023-09-04
Handle wait for result concerns

raman325:
Handles firs three bullets of https://github.com/home-assistant-libs/zwave-js-server-python/issues/738

### DIFF
--- a/zwave_js_server/model/node/__init__.py
+++ b/zwave_js_server/model/node/__init__.py
@@ -621,7 +621,7 @@ class Node(EventBase):
 
     async def async_abort_firmware_update(self) -> None:
         """Send abortFirmwareUpdate command to Node."""
-        await self.async_send_command("abort_firmware_update", wait_for_result=True)
+        await self.async_send_command("abort_firmware_update", wait_for_result=False)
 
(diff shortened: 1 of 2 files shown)


zwave_js_server/model/node/__init__.py
```diff
@@ -860,10 +862,11 @@ async def async_set_date_and_time(self, datetime_: datetime | None = None) -> bo
             "set_date_and_time",
             **args,
             require_schema=28,
-            wait_for_result=True,
+            wait_for_result=wait_for_result,
         )
-        assert data
-        return cast(bool, data["success"])
+        if data:
+            return cast(bool, data["success"])
```
MartinHjelmare, line 868: What's the difference between the success in the result and the success in the response message for this command?

MartinHjelmare:
Can we add a test for the uncovered code?

zwave_js_server/model/node/__init__.py
raman325: the success in the response message will be false if there is an exception thrown in the server (by the driver or by the server). The `success` in the result indicates whether the date and time was actually set.

MartinHjelmare (approved):

PR #754, opened 2023-09-11
Handle unknown and default Durations

raman325:
fixes https://github.com/home-assistant/core/issues/99873

### DIFF
--- a/zwave_js_server/model/duration.py
+++ b/zwave_js_server/model/duration.py
@@ -18,7 +18,7 @@ class Duration:
     """Duration class."""
 
     data: DurationDataType | Literal["unknown", "default"]
-    unit: Literal["seconds", "minutes"] = field(init=False)
+    unit: Literal["seconds", "minutes", "unknown", "default"] = field(init=False)
     value: int | float | None = field(init=False)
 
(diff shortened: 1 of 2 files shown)


MartinHjelmare (commented):
Black complains.

raman325:
> Black complains.

thanks, missed this

zwave_js_server/model/duration.py
```diff
@@ -17,12 +17,16 @@ class DurationDataType(TypedDict, total=False):
 class Duration:
     """Duration class."""
 
-    data: DurationDataType
+    data: DurationDataType | Literal["unknown", "default"]
```
MartinHjelmare, line 20: Does this mean that the unit in `DurationDataType` can only have seconds or minutes as values?

raman325: yup good call

raman325: reference: https://github.com/zwave-js/node-zwave-js/blob/master/packages/core/src/values/Duration.ts#L134

MartinHjelmare (approved):
Thanks!

PR #756, opened 2023-09-16
Handle non existing result futures on pop

raman325:
I'm not sure how to test this... My only thought at the moment is to patch `self._result_futures` with a dict subclass instance that overrides adding an item to the dict to not do it. But that seems like a bad hack, so open to other suggestions

### DIFF
--- a/test/test_client.py
+++ b/test/test_client.py
@@ -474,5 +474,5 @@ async def test_pop_future_none(client_session, url, driver_ready):
 
     await driver_ready.wait()
 
-    with pytest.raises(NotConnected):
+    with pytest.raises(asyncio.CancelledError):
         await client.async_send_command({"command": "some_command"})
--- a/test/test_client.py
+++ b/test/test_client.py
@@ -472,5 +472,7 @@ async def test_pop_future_none(client_session, url, driver_ready):
 
     asyncio.create_task(client.listen(driver_ready))
 
-    with pytest.raises(asyncio.CancelledError):
+    await driver_ready.wait()
+
+    with pytest.raises(NotConnected):
         await client.async_send_command({"command": "some_command"})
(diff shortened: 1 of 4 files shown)


MartinHjelmare:
I think we should try to test this by sending a command and waiting for the result but before that arrives we receive a websocket message with a message type that closes the connection. That will cancel all the result futures and clear the result futures dict.

I think we already patch out the WS client `send_json` method so that when called it will set the WS client as closed which will trigger the canceling and clearing that we want. I'd try to create a client, connect it, then create a listen task and wait for it to be ready, then send a command without mocking out the command first. That should trigger the closing of the WS client, and then a CancelledError will be raised when awaiting the sending of the command.

test/test_client.py
```diff
@@ -461,3 +461,16 @@ async def test_additional_user_agent_components(client_session, url):
...
+async def test_pop_future_none(client_session, url, driver_ready):
+    """Test when a future has been cleared from futures dict, popping still works."""
+    client = Client(url, client_session)
+    await client.connect()
+
+    assert client.connected
+
+    asyncio.create_task(client.listen(driver_ready))
+
+    with pytest.raises(asyncio.CancelledError):
```
MartinHjelmare, line 475: Should we await `driver_ready` first?

raman325: by awaiting `driver_ready` the connection is already closed by the time we try to send the command. Will have to review the fixture to see what the right implementation is here

raman325: the problem is that as soon as we drain the message queue, we close the connection. We should instead only close the connection if we attempt to receive a message and the queue is already drained. Making some adjustments to the tests accordingly

raman325: let me know what you think of these changes. I think this is the correct behavior, but I am not 100% sure

MartinHjelmare (approved):
Looks good!

PR #766, opened 2023-09-30
Filter out logging events from debug logs

raman325:
When server logging is enabled, we should not include the `Received message` logging statement for logging events because it's not formatted well and we'll get the formatted version immediately after anyway.

### DIFF
--- a/zwave_js_server/model/node/firmware.py
+++ b/zwave_js_server/model/node/firmware.py
@@ -297,7 +297,7 @@ class NodeFirmwareUpdateInfo:
         return cls(
             version=data["version"],
             changelog=data["changelog"],
-            channel=data.get("channel", "stable"),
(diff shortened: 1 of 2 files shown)


zwave_js_server/model/node/firmware.py
```diff
@@ -297,7 +297,7 @@ def from_dict(
         return cls(
             version=data["version"],
             changelog=data["changelog"],
-            channel=data["channel"],
+            channel=data.get("channel", "stable"),
```
MartinHjelmare, line 300: This doesn't match the PR title.

MartinHjelmare (changes requested):
Let's break out the change of the firmware model.

MartinHjelmare (approved):

PR #773, opened 2023-10-04
Create script to generate notification enums and check if git is dirty

raman325:
This is to prep for https://github.com/zwave-js/certification-backlog/issues/9

EDIT: I've also added a dirty repo check at the bottom of each script. The thought is that we can set up a workflow to run those once a day via tox. That will allow us to catch if/when the constants change


raman325:
may rework this to generate it like we do for multilevel sensor constants

marcelveldt (approved):
looks fine by me.
if we expect this to be ever going to change/extend, we might want to add a fallback handler to unknown ?

example: https://github.com/home-assistant-libs/aiohue/blob/master/aiohue/v2/models/feature.py#L97

raman325:
Need to rework enums to include the second key in the enum value names, also need to add state enums

scripts/generate_notification_constants.py
```diff
@@ -0,0 +1,315 @@
...
+
+import requests
+from slugify import slugify
+
+GITHUB_PROJECT = "zwave-js/node-zwave-js"
+BRANCH_NAME = "master"
+NOTIFICATIONS_FILE_PATH = "packages/config/config/notifications.json"
+
+CONST_FILE_PATH = pathlib.Path(__file__).parent.joinpath(
+    "../zwave_js_server/const/command_class/notification.py"
```
MartinHjelmare, line 19: Maybe use `parent` twice and remove the double dot?

MartinHjelmare (approved):

raman325:
this needs to be merged before we finalize the notification CC constants in this PR: https://github.com/zwave-js/node-zwave-js/pull/6368

PR #782, opened 2023-10-16
Set up pre-commit

raman325:
Starting to get annoyed with missing some of these checks before committing. Shamelessly stolen from `home-assistant/core`


.pre-commit-config.yaml
```diff
@@ -0,0 +1,47 @@
...
+      # committing from a GUI tool that was not launched from an activated
+      # shell.
+      - id: mypy
+        name: mypy
+        entry: scripts/run-in-env.sh mypy
+        language: script
+        types: [python]
+        require_serial: true
+        files: ^(zwave_js_server)/.+\.py$
+      - id: flake8
```
MartinHjelmare, line 28: Why do we need to run flake8 locally?

raman325: I'm mirroring what we do in CI for testing: https://github.com/home-assistant-libs/zwave-js-server-python/blob/4c41c10328d69b30c2ac72d8e478baa073ce024b/tox.ini#L16-L23

MartinHjelmare: Right but maybe we can run flake8 via a dedicated pre-commit hook? If there are special requirements on the environment a local hook could be needed. I'm not sure what the case is for flake8.

Side note: We should move to ruff instead of flake8 longterm.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,39 @@
...
+        language: script
+        types: [python]
+        require_serial: true
+        files: ^(zwave_js_server)/.+\.py$
+      - id: pylint
+        name: pylint
+        entry: scripts/run-in-env.sh pylint -j 0 --ignore-missing-annotations=y
+        language: script
+        types: [python]
+        files: ^zwave_js_server/.+\.py$
```
MartinHjelmare, line 39: No end of file line.

MartinHjelmare (approved):

PR #783, opened 2023-10-16
Add isort to pre-commit

raman325:


.pre-commit-config.yaml
```diff
@@ -6,6 +6,10 @@ repos:
         args:
           - --quiet
         files: ^((zwave_js_server|scripts|test)/.+)?[^/]+\.py$
+  - repo: https://github.com/pycqa/isort
```
MartinHjelmare, line 9: I'd move to ruff directly instead.

raman325: sounds good

PR #786, opened 2023-10-16
Switch to ruff in tests and precommit

raman325:
New to `ruff` so if you wonder why I did something a certain way, I probably didn't have a good reason, I was just experimenting/using HA as a reference while trying to minimize complexity. I fixed some things that ruff caught so that's great.

~~I disabled `warn_return_any` because all of a sudden it started spitting out a bunch of errors. Not sure why we didn't see these before? If we want to re-enable it, I can do so in a separate PR, just let me know~~ I re-enabled it without issue and ran `mypy` to confirm there were no errors, so not sure what happened


.pre-commit-config.yaml
```diff
@@ -1,4 +1,18 @@
...
+        name: Check ruff for zwave_js_server
+        files: ^(zwave_js_server)/.+\.py$
+        args: [--fix, --exit-non-zero-on-fix]
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    rev: v0.0.292
+    hooks:
+      - id: ruff
+        name: Check ruff for scripts and test
+        files: ^(scripts|test)/.+\.py$
+        args: [--fix, --exit-non-zero-on-fix, --select, I001, D210, D400, D403, D415]
```
MartinHjelmare, line 15: What's the difference between these two hooks? Are we more forgiving for the tests and scripts?

PR #791, opened 2023-10-22
Add lock setConfiguration util cmd

raman325:
This new util command can be used to set the lock's configuration. We are planning to make this a service call for locks.

@AlCalzone I know Python isn't your thing but take a skim and let me know if anything looks off from what we discussed in https://github.com/zwave-js/certification-backlog/issues/21

### DIFF
--- a/zwave_js_server/const/command_class/lock.py
+++ b/zwave_js_server/const/command_class/lock.py
@@ -5,8 +5,11 @@ Includes Door Lock and Lock CCs.
 """
 from __future__ import annotations
 
+from dataclasses import dataclass, field
 from enum import Enum, IntEnum
+from typing import TypedDict
 
+from ...model.endpoint import Endpoint
 from .. import CommandClass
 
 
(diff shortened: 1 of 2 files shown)


zwave_js_server/util/lock.py
```diff
@@ -144,3 +151,58 @@ async def clear_usercode(node: Node, code_slot: int) -> SetValueResult | None:
...
+                value
+                for value in endpoint.values.values()
+                if value.command_class == CommandClass.DOOR_LOCK
+                and value.property_name == property_name
+            ),
+            None,
+        )
+        if (
+            val := getattr(configuration, attr_name)
+        ) is not None and cached_value is None:
```
MartinHjelmare, line 189: Start with the check of `cached_value` as that will be the fastest. We'll then avoid the `getattr` if the first check is false.

zwave_js_server/const/command_class/lock.py
```diff
@@ -128,3 +131,68 @@ class CodeSlotStatus(IntEnum):
...
+    def to_dict(self) -> DoorLockCCConfigurationSetOptionsDataType:
+        """Convert command options to dict."""
+        data = DoorLockCCConfigurationSetOptionsDataType(
+            operationType=self.operation_type,
+            outsideHandlesCanOpenDoorConfiguration=(
+                self.outside_handles_can_open_door_configuration
+            ),
+            insideHandlesCanOpenDoorConfiguration=(
+                self.inside_handles_can_open_door_configuration
+            ),
```
AlCalzone, line 187: Not 100% sure about this, my python is very rusty.
If `self.outside_handles_can_open_door_configuration` is a tuple, doesn't this result in `outsideHandlesCanOpenDoorConfiguration` being a nested tuple?

AlCalzone (commented):
Seems to match what we discussed

zwave_js_server/const/command_class/lock.py
MartinHjelmare: The parenthesis here is just to be able to write the statement on more than one line.

A tuple with a single item is written like this:

```py
is_tuple = (1, )

```
This is not a tuple:
```py
not_tuple = (1)
```

zwave_js_server/util/lock.py
raman325: but we need to know the `val` in the next check, and we don't want to retrieve it twice right?

MartinHjelmare: Ok

MartinHjelmare (approved):

MartinHjelmare (approved):

PR #801, opened 2023-10-27
Handle reserved thermostat setpoints

raman325:
From the ensuing conversations about https://github.com/home-assistant/core/issues/102637, I think we need fallbacks for these enums so that we can handle these issues gracefully in the core but using some of the existing mechanisms like converting values to enums immediately

### DIFF
--- a/zwave_js_server/const/command_class/thermostat.py
+++ b/zwave_js_server/const/command_class/thermostat.py
@@ -22,6 +22,7 @@ class ThermostatMode(IntEnum):
     """Enum with all (known/used) Z-Wave ThermostatModes."""
 
     # https://github.com/zwave-js/node-zwave-js/blob/master/packages/zwave-js/src/lib/commandclass/ThermostatModeCC.ts#L53-L70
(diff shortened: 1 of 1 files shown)


zwave_js_server/const/command_class/thermostat.py
```diff
@@ -75,6 +89,14 @@ class ThermostatSetpointType(IntEnum):
     AWAY_COOLING = 14
     FULL_POWER = 15
 
+    @classmethod
+    def _missing_(cls: type, value: object) -> ThermostatSetpointType:
+        """Set default enum member if an unknown value is provided."""
+        # Handle reserved values
+        if isinstance(value, int) and 3 <= value <= 6:
```
MartinHjelmare, line 96: What has reserved 3-6?

raman325: It's reserved in the spec, I guess it would be device specific? To kpine's point below, this solution doesn't work, will close and open a new one when ready

PR #815, opened 2023-11-14
Add orjson support

### LINKED ISSUE #814: TODO: Use `orjson` if it's available

raman325:
fixes #814

Unclear how I would test this, open to suggestions


zwave_js_server/client.py
```diff
@@ -39,6 +40,11 @@
 from .model.log_message import LogMessage
 from .model.version import VersionInfo, VersionInfoDataType
 
+try:
+    import orjson as json
```
MartinHjelmare, line 44: Let's add this as an `extra` dependency that library users can select when installing the library.

MartinHjelmare (changes requested):

MartinHjelmare (approved):

MartinHjelmare:
We could add an additional tox job that installs the orjson requirement before running the tests.

PR #831, opened 2023-12-07
Add "Soil temperature" property for thermostat

TastyPi:
The HELTUN HE-HT01 thermostat provides both an "Air temperature" and "Soil temperature" property in its `CommandClass.SENSOR_MULTILEVEL`. I'd like to create a `ZWaveDiscoverySchema` which displays the correct temperature based on the "Source Sensor" configuration parameter ([parameter number 11](https://devices.zwave-js.io/?jumpTo=0x0344:0x0004:0x0001:0.0)).

Other code in `home-assistant/core` uses constants from this lib to refer to properties, so I assume it is best practice to list them here before using them there. Let me know if I'm mistaken.


raman325:
hey so I went ahead and implemented this because these scripts are not architected very well and it would take you a lot longer than it just took me to make the necessary changes. Thank you again though, this was a great idea

PR #836, opened 2023-12-09
Add CI script to run auto-gen scripts daily

raman325:
This will run the auto-gen scripts every morning and open a PR if there are new changes to commit

Will merge after #835


.github/workflows/check-generated-modules.yml
```diff
@@ -0,0 +1,40 @@
...
+      - name: Check for changes
+        id: changes
+        uses: UnicornGlobal/has-changes-action@v1.0.12
+      - name: Create Pull Request
+        if: steps.changes.outputs.changed == 1
+        uses: peter-evans/create-pull-request@v5.0.2
+        with:
+          commit-message: "[create-pull-request] Update auto-generated constants"
+          delete-branch: true
+          title: "[create-pull-request] Update auto-generated constants"
```
MartinHjelmare, line 40: What does the `[create-pull-request]` mean? I'd remove that.

raman325: it's the name of the action generating the PR. Do you have a suggestion for what to prepend it with? Just want to make it obvious that these are not user generated PRs

MartinHjelmare: Github actions should be the user that opens the PR. It's pretty clear that it's not user generated. We don't need to prepend anything to the PR title.

Example:
https://github.com/home-assistant-libs/pytradfri/pull/787

raman325: ok fair enough

MartinHjelmare (approved):

PR #843, opened 2023-12-14
Drop min schema version required

raman325:
We technically don't need to support the latest schema because worst case scenario, we get nothing back for existing rebuild route progress state. If we do the dependency bump to this, it's no longer a breaking change

### DIFF
--- a/zwave_js_server/const/__init__.py
+++ b/zwave_js_server/const/__init__.py
@@ -10,7 +10,7 @@ PACKAGE_NAME = "zwave-js-server-python"
 __version__ = "0.55.0"
 
 # minimal server schema version we can handle
-MIN_SERVER_SCHEMA_VERSION = 34
+MIN_SERVER_SCHEMA_VERSION = 33
 # max server schema version we can handle (and our code is compatible with)
 MAX_SERVER_SCHEMA_VERSION = 34
 


MartinHjelmare:
How will we keep track of the compatibility of different server versions and the client version?

raman325:
Are the constants not sufficient? We only need to bump the minimum if a model changes or if there is new key data we are relying on

MartinHjelmare:
Personally I won't be able to remember what the consequences are of running a server of version 33 with a client that has support for version 34 if we don't have some system for tracking that.

> we get nothing back for existing rebuild route progress state

How will this affect the user?

raman325:
the point of pulling this data in as part of the state dump was so that if a user pulled up the rebuild routes modal after triggering it from elsewhere (e.g. zwave-js-ui), we could immediately display progress. Without it, progress is delayed until the next progress event gets emitted.

raman325:
> Personally I won't be able to remember what the consequences are of running a server of version 33 with a client that has support for version 34 if we don't have some system for tracking that.

We had a doc that tracked schema changes in the server repo but it's long been abandoned. I can create a task to refresh it up to latest, it will just take a while to parse through all of the changes since the last time it was updated

MartinHjelmare:
History isn't that imported. It's the current possible compatibility matrix that is interesting. If a user encounters an issue we're always going to ask them to update to the latest core version. We can skip the previous versions before schema 33 if it's a lot of work.

raman325:
> History isn't that imported. It's the current possible compatibility matrix that is interesting. If a user encounters an issue we're always going to ask them to update to the latest core version. We can skip the previous versions before schema 33 if it's a lot of work.

Sorry not following. The server supports all schemas, the issue is the library and the integration. In this case, we can support schema 33 with unimproved functionality but it will still work. So I'm not sure what the action is to take, can you clarify?

MartinHjelmare:
I was responding to this:

> I can create a task to refresh it up to latest, it will just take a while to parse through all of the changes since the last time it was updated

raman325:
gotcha I meant documenting all of the schema changes here: https://github.com/zwave-js/zwave-js-server/blob/master/API_SCHEMA.md

MartinHjelmare:
Yes. I'm suggesting jumping to schema 33 documenting that and doing the older ones later as time permits. The important information right now is schema 33 and 34.

raman325:
@MartinHjelmare are we good to merge here then?

MartinHjelmare (approved):
Thanks!

PR #856, opened 2023-12-27
Revert "Add orjson support (#815)"

MartinHjelmare:
- This reverts commit 392543e38901be306dc4cd266c1487d608fedd09.
- We allowed orjson instead of regular json for serialization. But orjson returns bytes instead of string when serializing, which isn't compatible with the websocket connection.
- Revert for now until we have time to fix it.


raman325 (approved):
thanks for picking this up

MartinHjelmare:
Verified working locally. :+1:

PR #860, opened 2024-01-05
Add exception handling to EventBase.emit

raman325:
Prevents the lib from raising an exception while handling callbacks for events received from the server


MartinHjelmare (approved):
Thanks!

PR #911, opened 2024-02-29
Get last seen even if its not part of statistics

### LINKED ISSUE #909: "last seen" statistic is not always available
A "last seen" sensor was added to HA to allow users to track that last time the node was involved in any communications. It tracks the "last seen" node statistic field. However, these statistics are not persisted to cache and are lost on a Z-Wave JS restart. Therefore the sensor will go "Unknown" after a restart. Mains-powered nodes are pinged on startup, so the sensor will be corrected quickly, but battery nodes won't update until they go awake or send some message. This is most obvious in dead nodes (never send updates) or battery devices that never wake or are rarely used.

Instead, there is the [`node.last_seen`](https://github.com/home-assistant-libs/zwave-js-server-python/blob/d050e6961c86076dc49fd0ac3242c3252b4b1d07/zwave_js_server/model/node/__init__.py#L360) property. This can be used in some way (either here or in the integration) to seed the value until the statistic is available.

cc @raman325

raman325:
fixes #909

### DIFF
--- a/test/model/test_node.py
+++ b/test/model/test_node.py
@@ -189,6 +189,13 @@ def test_from_state(client):
     assert node.last_seen == datetime(2023, 7, 18, 15, 42, 34, 701000, UTC)
 
 
+async def test_last_seen(lock_schlage_be469):
+    """Test last seen property."""
+    assert lock_schlage_be469.last_seen == datetime(
+        2023, 7, 18, 15, 42, 34, 701000, UTC
+    )
+
+
 async def test_highest_security_value(lock_schlage_be469, ring_keypad):
     """Test the highest_security_class property."""
     assert lock_schlage_be469.highest_security_class == SecurityClass.S0_LEGACY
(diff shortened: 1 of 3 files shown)


pumrum:
thank you so much for working on this - can't wait to see it pushed to the live version so I can check it out!

test/model/test_node.py
```diff
@@ -189,6 +189,13 @@ def test_from_state(client):
     assert node.last_seen == datetime(2023, 7, 18, 15, 42, 34, 701000, UTC)
 
 
+async def test_last_seen(lock_schlage_be469):
+    """Test last seen property."""
+    assert lock_schlage_be469.last_seen == datetime(
```
MartinHjelmare, line 194: Can we also test the statistics? Ie that there's this attribute in the node statistics even if the statistics data is missing this item.

MartinHjelmare (approved):
Thanks!

PR #974, opened 2024-05-14
Fix QRProvisioningInformation.from_dict()

raman325:
Currently the logic returns an empty list when supported protocols don't exist, but it should be `None`. There's a similar problem with additional properties


raman325:
Nothing is wrong with it, but if we take what we get back from the server and do `QRProvisioningInformation.from_dict(data).to_dict()` we should get the same output as we had for input right? The empty list makes that not true, it should be None or null

MartinHjelmare (approved):
Looks good!

PR #996, opened 2024-06-14
Update nodes internal state data

raman325:
When making updates to the lib's last seen state, we were not updating its internal state (`node.data`). The sensor entity uses `node.data` to populate its state, and if there is no last seen state data, the entity becomes `unknown` as a result.

I'm separately wondering if we should update the statistics sensor entity descriptions to use the actual model (e.g. `node.last_seen`) instead of the node's internal state (`node.data["lastSeen"]`)). Do we generally want to avoid accessing the internal state of things in the lib from the integration code? If so, I'm guessing there are other places we are doing that as well.

Thanks @kpine for finding the issue!


MartinHjelmare (approved):
Yes, we should use the appropriate model properties / attributes.

PR #1051, opened 2024-09-05
Fix missing lock timeout configuration

MartinHjelmare:
- This PR will solve the not working door lock timeout configuration parameter documented in this issue:
https://github.com/zwave-js/certification-backlog/issues/21
- The parameter existed in the dataclass but we had missed to include it when converting the dataclass to its dict representation, which is used when sending the command to the server.

### DIFF
--- a/test/common.py
+++ b/test/common.py
@@ -1,7 +1,17 @@
 """Common methods for tests."""
 
 import asyncio
+from typing import Protocol
 from unittest.mock import AsyncMock
+
+
+class MockCommandProtocol(Protocol):
+    """Represent the function signature of mock_the mock_command callable."""
+
+    def __call__(
+        self, command: dict, response: dict, success: bool = True
+    ) -> list[dict]:
+        """Represent the signature of the mock_command callable."""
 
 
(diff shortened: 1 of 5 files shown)


test/common.py
```diff
@@ -1,9 +1,19 @@
 """Common methods for tests."""
 
 import asyncio
+from typing import Protocol
 from unittest.mock import AsyncMock
 
 
+class MockCommandProtocol(Protocol):
```
MartinHjelmare, line 8: I added some type annotations for the new test, to not have missing type information.

test/common.py
```diff
@@ -1,9 +1,19 @@
...
+    """Represent the function signature of mock_the mock_command callable."""
```
MartinHjelmare, line 9:
```suggestion
    """Represent the function signature of the mock_command callable."""
```

marcelveldt (approved):
Looks good! I'm still not convinced that this should be a service/action (rather something hidden in expert panel) but that discussion can be held later. At least now its fixed!

PR #1073, opened 2024-09-29
Fix inclusion started event model

MartinHjelmare:
- We missed updating the inclusion started event model when updating the library to handle changes in Z-Wave JS v 13.x. Fix that!
- Users of the `InclusionStartedEventModel` model need to update their attribute access to use the new attribute `strategy` instead of `secure` which was removed in v 13 of Z-Wave JS.


raman325 (approved):
thanks, missed this but confirmed the change was made upstream in the last release

PR #1079, opened 2024-10-03
Update const generation scripts based on zwave-js 13

raman325:
The existing scripts to generate the constants for the Notification and Multilevel Sensor CCs depended on JSON files that have been retired in Z-Wave JS 13.x in favor of in-code constants.

To access these constants so we can serialize the data so it can be parsed by the Python scripts, we now run a Node script first.

Looks like I also fixed a sorting issue in the output


zwave_js_server/const/command_class/notification.py
```diff
@@ -149,10 +166,10 @@ def _missing_(
 class BarrierVacationModeNotificationEventValue(NotificationEventValue):
     """Enum for known barrier vacation mode notification event value."""
 
-    # https://github.com/zwave-js/node-zwave-js/blob/master/packages/config/config/notifications.json
+    # https://github.com/zwave-js/node-zwave-js/blob/master/packages/core/src/registries/Notifications.ts
     UNKNOWN = -1
     MODE_DISABLED = 0
-    MODE_ENABLED = 255
+    MODE_ENABLED = 597
```
MartinHjelmare, line 172: Is this correct?

MartinHjelmare (changes requested):

zwave_js_server/const/command_class/notification.py
raman325: no, good catch

scripts/generate_multilevel_sensor_constants.py
```diff
@@ -37,15 +36,15 @@ def normalize_scale_definition(scale_definitions: dict[str, dict]) -> dict[str,
 scales = {}
 sensors = {}
 
-for sensor_props in get_json_file("sensors.json"):
+for sensor_props in pathlib.Path("sensors.json").read_text():
```
MartinHjelmare, line 39: We still need to load the json to native types.

MartinHjelmare (approved):
Thanks!

Issue #925, opened 2024-03-12
async_set_raw_config_parameter_value doesn't allow setting of "raw" config parameters


kpine:
`Endpoint.async_set_raw_config_parameter_value` is supposed to allow setting of "raw" config parameters, which are parameters that are not known to Z-Wave JS, using the CC API instead of Value API. Users can provide 1) parameter number, 2) value, 3) value size, 4) value format and Z-Wave JS will write the value. However, this function actually does some [validation](https://github.com/home-assistant-libs/zwave-js-server-python/blob/e939722e29b6f11d1ab97b2459a1034367df669a/zwave_js_server/model/endpoint.py#L301-L334) and prevents this use case.

The function should not need property and property key arguments, and should just call the CC API without validation (except for perhaps validating the parameters are defined).

PR #1080, opened 2024-10-03
Remove validation for async_set_raw_config_parameter_value

raman325:
## Breaking change

The return result for `async_set_raw_config_parameter_value` has been changed from a tuple that includes the targeted Z-Wave value to just the result

resolves https://github.com/home-assistant-libs/zwave-js-server-python/issues/925


MartinHjelmare (approved):
Thanks!

Let's wait a week before we merge this so it's easier to release patch releases if we find a bug in the Core October release.

PR #1089, opened 2024-10-10
Add lock util function to set multiple usercodes

raman325:
Z-Wave JS allows applications to send locks a single command with multiple user codes to set. This PR adds a utility function so that we can create a corresponding service in Home Assistant

### DIFF
--- a/test/util/test_lock.py
+++ b/test/util/test_lock.py
@@ -20,6 +20,7 @@ from zwave_js_server.const.command_class.lock import (
 )
 from zwave_js_server.exceptions import NotFoundError
 from zwave_js_server.model.node import Node
+from zwave_js_server.model.value import SupervisionResult
 from zwave_js_server.util.lock import (
     clear_usercode,
     get_code_slots,
@@ -116,7 +117,9 @@ async def test_set_usercodes(lock_schlage_be469, mock_command, uuid4):
     )
 
     # Test wrong types to ensure values get converted
-    await set_usercodes(node, {"1": 1234})
+    assert await set_usercodes(node, {"1": 1234}) == SupervisionResult(
+        {"status": SupervisionStatus.SUCCESS}
+    )
     assert len(ack_commands) == 1
     assert ack_commands[0] == {
         "command": "endpoint.invoke_cc_api",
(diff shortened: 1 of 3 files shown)


test/util/test_lock.py
```diff
@@ -102,6 +108,57 @@ async def test_set_usercode(lock_schlage_be469, mock_command, uuid4):
     assert len(ack_commands) == 1
 
 
+async def test_set_usercodes(lock_schlage_be469, mock_command, uuid4):
```
MartinHjelmare, line 111: Let's add type annotations for new tests.
```suggestion
async def test_set_usercodes(
    lock_schlage_be469: Node, mock_command: MockCommandProtocol, uuid4: str
) -> None:
```

test/util/test_lock.py
```diff
@@ -102,6 +108,57 @@ async def test_set_usercode(lock_schlage_be469, mock_command, uuid4):
...
+
+    # Test invalid code length
+    with pytest.raises(ValueError):
+        await set_usercodes(node, {1: "123"})
+
+    # assert no new command calls
+    assert len(ack_commands) == 1
+
+
+async def test_set_usercodes_invalid(lock_schlage_be469, mock_command):
```
MartinHjelmare, line 150:
```suggestion
async def test_set_usercodes_invalid(
    lock_schlage_be469: Node, mock_command: MockCommandProtocol
) -> None:
```

zwave_js_server/util/lock.py
```diff
@@ -141,13 +143,39 @@ async def set_usercode(
...
+    if any(len(str(usercode)) < 4 for usercode in codes.values()):
+        raise ValueError("User codes must be at least 4 digits")
+
+    cc_api_codes = [
+        {
+            LOCK_USERCODE_ID_PROPERTY: int(code_slot),
+            LOCK_USERCODE_STATUS_PROPERTY: CodeSlotStatus.ENABLED,
+            LOCK_USERCODE_PROPERTY: str(usercode),
+        }
+        for code_slot, usercode in codes.items()
```
MartinHjelmare, line 165: Can we avoid iterating the codes twice?

MartinHjelmare (changes requested):

MartinHjelmare (approved):
Thanks!

Issue #1091, opened 2024-10-10
TODO: Remove redundant validation


raman325:
`Z-Wave JS` does a lot of internal validation that we have attempted to replicate within the library, and in some cases we may have been overly strict. Where it makes sense, we should fallback to the driver validation. To accomplish this we will need to make sure that whatever cases that lead to an exception also would return an error from zwave-js are removed from the library.

CC @MartinHjelmare as this is something we have tackled here or there, but I think this is an overall design decision we should align on

PR #1092, opened 2024-10-12
Remove validation from lock util functions

raman325:
For #1091

### DIFF
--- a/zwave_js_server/util/lock.py
+++ b/zwave_js_server/util/lock.py
@@ -143,12 +143,7 @@ async def set_usercode(
 ) -> SetValueResult | None:
     """Set the usercode to index X on the lock."""
     value = get_code_slot_value(node, code_slot, LOCK_USERCODE_PROPERTY)
-    usercode = str(usercode)
-
-    if len(usercode) < 4:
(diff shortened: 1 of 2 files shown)


zwave_js_server/util/lock.py
```diff
@@ -143,12 +143,7 @@ async def set_usercode(
 ) -> SetValueResult | None:
     """Set the usercode to index X on the lock."""
     value = get_code_slot_value(node, code_slot, LOCK_USERCODE_PROPERTY)
-    usercode = str(usercode)
-
-    if len(usercode) < 4:
-        raise ValueError("User code must be at least 4 digits")
-
-    return await node.async_set_value(value, usercode)
+    return await node.async_set_value(value, str(usercode))
```
MartinHjelmare, line 146: Do we need to copy the usercode to string? The parameter is already typed as a string.

raman325: this may have been a workaround to address upstream issues. Typing is right, and if there are issues they should probably be fixed upstream

MartinHjelmare (approved):
Thanks!

PR #1117, opened 2024-11-13
Fix message format for get/set raw config

MindFreeze:
Format issue discovered by @AlCalzone https://github.com/zwave-js/certification-backlog/issues/31#issuecomment-2470432220
I am confused as to how `async_set_raw_config_parameter_value` worked before but maybe it didn't 🤷‍♂️

### DIFF
--- a/zwave_js_server/model/endpoint.py
+++ b/zwave_js_server/model/endpoint.py
@@ -324,7 +324,7 @@ class Endpoint(EventBase):
 
         data = await self.async_send_command(
             "set_raw_config_parameter_value",
-            options={k: v for k, v in options.items() if v is not None},
+            **options,
             require_schema=33,
         )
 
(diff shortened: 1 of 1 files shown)


zwave_js_server/model/endpoint.py
```diff
@@ -324,7 +324,7 @@ async def async_set_raw_config_parameter_value(
 
         data = await self.async_send_command(
             "set_raw_config_parameter_value",
-            options={k: v for k, v in options.items() if v is not None},
+            **options,
```
MartinHjelmare, line 327: Is it ok to remove the filtering of options set to `None` or should we keep that filter before unpacking the options?

AlCalzone: I think it needs to be filtered first

AlCalzone:
> I am confused as to how `async_set_raw_config_parameter_value` worked before but maybe it didn't 🤷‍♂️

It didn't - at least in my test.

MartinHjelmare (approved):
Thanks!

PR #1153, opened 2025-01-15
Add EU Long Range to RFRegion

AlCalzone:


MartinHjelmare (approved):
Thanks!

Do you know what the correct link should be for line 323 nowadays?

AlCalzone:
https://github.com/zwave-js/node-zwave-js/blob/master/packages/core/src/definitions/RFRegion.ts#L1

MartinHjelmare:
Great!

PR #1172, opened 2025-02-18
Add method to update the controller's inclusion state

AlCalzone:
This is necessary to be able to handle the `inclusion state changed` event, which is currently being ignored by HA, leading to a discrepancy between the assumed and actual inclusion state.

I'm not sure if this is the right approach. Ideally I'd like a method to update any property, not just the inclusion state, so we don't have to add another when we need to selectively update a different property in the future.

### DIFF
--- a/zwave_js_server/model/controller/__init__.py
+++ b/zwave_js_server/model/controller/__init__.py
@@ -261,6 +261,10 @@ class Controller(EventBase):
                 self.data["rebuildRoutesProgress"]
             )
 
+    def update_inclusion_state(self, state: InclusionState) -> None:
+        """Update inclusion state."""
+        self.data["inclusionState"] = state
+
     async def async_begin_inclusion(
         self,
         inclusion_strategy: Literal[


zwave_js_server/model/controller/__init__.py
```diff
@@ -261,6 +261,10 @@ def update(self, data: ControllerDataType) -> None:
                 self.data["rebuildRoutesProgress"]
             )
 
+    def update_inclusion_state(self, state: InclusionState) -> None:
```
MartinHjelmare, line 264: Update the event handler `handle_inclusion_state_changed` instead.

Each incoming event should have a corresponding event handler in the appropriate model. The event handler should update state as needed before emitting the event, calling event listeners of the event. This is how the library is designed.

MartinHjelmare (changes requested):

zwave_js_server/model/controller/__init__.py
MindFreeze: https://github.com/home-assistant-libs/zwave-js-server-python/pull/1173

AlCalzone:
Closing in favor of #1173. That makes more sense.

PR #1173, opened 2025-02-18
Update controller inclusion state in `handle_inclusion_state_changed`

MindFreeze:
New version of #1172 
@MartinHjelmare This is what you mean right?


MartinHjelmare (commented):
Yes, thanks! Please add a test that checks that the state is updated after this event is received.

MartinHjelmare (approved):
Thanks!

PR #1182, opened 2025-03-04
Add function to backup/restore NVM in base64 format

MindFreeze:
This will allow us to send it to the frontend over the websocket without redundant encoding/decoding.
Was wondering whether to update the existing functions or add new ones 🤷‍♂️


MartinHjelmare (approved):
Thanks!

PR #1185, opened 2025-03-10
Make disable_server_logging a callback

MartinHjelmare:
- The `Client.disable_server_logging` method doesn't need to be a coroutine function since it doesn't do any I/O or blocking calls. The client doesn't need to be connected to call this method.
- Update affected test and add type annotations for it and used fixtures.


raman325 (approved):
thanks for the code cleanup <3

PR #1196, opened 2025-03-20
Use PEP 639 license expression

cdce8p:
Followup to https://github.com/home-assistant-libs/zwave-js-server-python/pull/1029

With v77 setuptools added full support for PEP 639 license expression. With that it's now possible to use `project.license` directly which will map to `License-Expression` in the project metadata.


MartinHjelmare (approved):
Thanks!

PR #1245, opened 2025-06-24
Add migrateOptions to nvm restore methods

MindFreeze:
For https://github.com/zwave-js/backlog/issues/88

Wasn't sure if I should increase the require_schema. The options were added in 42 but the command would work in earlier versions. The options would just be ignored.

### DIFF
--- a/zwave_js_server/model/controller/__init__.py
+++ b/zwave_js_server/model/controller/__init__.py
@@ -736,12 +736,13 @@ class Controller(EventBase):
         )
         return convert_base64_to_bytes(data["nvmData"])
 
-    async def async_restore_nvm(self, file: bytes) -> None:
+    async def async_restore_nvm(self, file: bytes, options: dict[str, bool] = {}) -> None:
         """Send restoreNVM command to Controller."""
         await self.client.async_send_command(
             {
                 "command": "controller.restore_nvm",
                 "nvmData": convert_bytes_to_base64(file),
+                "migrateOptions": options,
             },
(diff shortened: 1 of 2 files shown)


MartinHjelmare:
I think it's ok to require schema 42 now for it. It's less surprising.

zwave_js_server/model/controller/__init__.py
```diff
@@ -736,14 +736,15 @@ async def async_backup_nvm_raw(self) -> bytes:
         )
         return convert_base64_to_bytes(data["nvmData"])
 
-    async def async_restore_nvm(self, file: bytes) -> None:
+    async def async_restore_nvm(self, file: bytes, options: dict[str, bool] = {}) -> None:
```
MartinHjelmare, line 739: Don't set a mutable object as default argument. Set it to `None`.

zwave_js_server/model/controller/__init__.py
```diff
@@ -736,14 +736,15 @@ async def async_backup_nvm_raw(self) -> bytes:
...
         """Send restoreNVM command to Controller."""
         await self.client.async_send_command(
             {
                 "command": "controller.restore_nvm",
                 "nvmData": convert_bytes_to_base64(file),
+                "migrateOptions": options,
```
MartinHjelmare, line 745: Can we always pass this parameter even if there are no options set?

MartinHjelmare (changes requested):

PR #1250, opened 2025-07-11
Wait for node command completion when node is dead

AlCalzone:
When https://github.com/home-assistant/core/pull/148611 gets merged, we should also keep track of the command execution when a node is currently considered dead - it might not actually be. With this change, we return the command result instead of `None` in that case.

### DIFF
--- a/zwave_js_server/model/node/__init__.py
+++ b/zwave_js_server/model/node/__init__.py
@@ -505,7 +505,6 @@ class Node(EventBase):
             return result
         if wait_for_result is None and self.status not in (
             NodeStatus.ASLEEP,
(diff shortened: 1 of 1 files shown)


MartinHjelmare (commented):
We'll stop waiting for the result if the status changes to dead. Is that correct?

AlCalzone:
> We'll stop waiting for the result if the status changes to dead. Is that correct?

Z-Wave JS will reject all pending transactions for a node when it goes dead. So I would say it's unnecessary, but doesn't hurt to do.

MartinHjelmare (approved):

PR #1253, opened 2025-07-15
Drop support for pydantic v1

cdce8p:
Pydantic will only support Python 3.14 for v2 models. https://github.com/pydantic/pydantic/issues/11613#issuecomment-3061291322

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,7 +19,11 @@ classifiers = [
     "Programming Language :: Python :: 3.13",
     "Topic :: Home Automation",
 ]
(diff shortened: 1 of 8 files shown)


cdce8p:
> There is one more error I still need to look at.

Found the issue in the HA integration. It referenced `pydantic.v1.ValidationError` which should be changed to just `pydantic.ValidationError`. The full diff required for the next update https://github.com/home-assistant/core/commit/a65658d4d2c7839f91ccdf521792e98ce7cfd9a9

MartinHjelmare (approved):
Thanks!

PR #1260, opened 2025-07-24
Add driver firmware v15 commands

MartinHjelmare:
- Implement the changes in Z-Wave JS v15: https://zwave-js.github.io/zwave-js/#/getting-started/migrating/v15?id=moved-otw-firmware-update-functionality-from-controller-to-driver-class
- Bumped to schema version 44 to allow new signature of `async_firmware_update_otw`.

### DIFF
--- a/test/model/common.py
+++ b/test/model/common.py
@@ -1,17 +1,21 @@
 """Provide common tools for testing Z-Wave JS server models."""
 
-FIRMWARE_UPDATE_INFO = {
-    "version": "1.0.0",
-    "changelog": "changelog",
-    "channel": "stable",
-    "files": [{"target": 0, "url": "http://example.com", "integrity": "test"}],
-    "downgrade": True,
-    "normalizedVersion": "1.0.0",
-    "device": {
-        "manufacturerId": 1,
-        "productType": 2,
-        "productId": 3,
-        "firmwareVersion": "0.4.4",
-        "rfRegion": 1,
-    },
-}
+from zwave_js_server.model.firmware import FirmwareUpdateInfoDataType
+
+FIRMWARE_UPDATE_INFO = FirmwareUpdateInfoDataType(
+    **{
+        "version": "1.0.0",
+        "changelog": "changelog",
+        "channel": "stable",
+        "files": [{"target": 0, "url": "http://example.com", "integrity": "test"}],
+        "downgrade": True,
+        "normalizedVersion": "1.0.0",
+        "device": {
+            "manufacturerId": 1,
+            "productType": 2,
+            "productId": 3,
+            "firmwareVersion": "0.4.4",
+            "rfRegion": 1,
+        },
+    }
+)
--- a/test/model/common.py
+++ b/test/model/common.py
@@ -1,0 +1,17 @@
+"""Provide common tools for testing Z-Wave JS server models."""
+
+FIRMWARE_UPDATE_INFO = {
+    "version": "1.0.0",
+    "changelog": "changelog",
+    "channel": "stable",
+    "files": [{"target": 0, "url": "http://example.com", "integrity": "test"}],
+    "downgrade": True,
+    "normalizedVersion": "1.0.0",
+    "device": {
+        "manufacturerId": 1,
+        "productType": 2,
+        "productId": 3,
+        "firmwareVersion": "0.4.4",
+        "rfRegion": 1,
+    },
+}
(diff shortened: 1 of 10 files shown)


test/model/common.py
```diff
@@ -0,0 +1,17 @@
+"""Provide common tools for testing Z-Wave JS server models."""
+
+FIRMWARE_UPDATE_INFO = {
```
MartinHjelmare, line 3: This is a pure move.

zwave_js_server/firmware.py
```diff
@@ -44,7 +45,7 @@ async def update_firmware(
 
 async def driver_firmware_update_otw(
     url: str,
-    firmware_file: DriverFirmwareUpdateData,
+    firmware_file: FirmwareUpdateData,
```
MartinHjelmare, line 48: I've renamed this since the driver has all items that the node firmware updates has.

zwave_js_server/model/controller/__init__.py
```diff
@@ -876,13 +876,6 @@ async def async_firmware_update_ota(
         )
         return NodeFirmwareUpdateResult(node, data["result"])
 
-    async def async_is_firmware_update_in_progress(self) -> bool:
-        """Send isFirmwareUpdateInProgress command to Controller."""
-        data = await self.client.async_send_command(
-            {"command": "controller.is_firmware_update_in_progress"}, require_schema=26
-        )
-        return cast(bool, data["progress"])
-
```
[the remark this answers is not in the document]
MartinHjelmare: We require minimum schema version 41 before this PR in the client library. It has been removed in that version of the schema, right?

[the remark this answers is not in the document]
MartinHjelmare: Or maybe it's still supported by the server even on that schema?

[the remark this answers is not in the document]
MartinHjelmare: I can put it back if it's still supported.

MartinHjelmare:
I'll finish this tomorrow. Please let me know what you think on the open questions above.

zwave_js_server/model/controller/__init__.py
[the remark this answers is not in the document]
AlCalzone: I opted for backwards compatibility when making the changes:
https://github.com/zwave-js/zwave-js-server/blob/fd658faf9ea28bed786c54bcf5af401e4755bbdd/src/lib/controller/message_handler.ts#L379-L382

zwave_js_server/firmware.py
AlCalzone: I have no plans for this to diverge, as it would require changes to the update service for no significant benefit.

MartinHjelmare: I think I should make a wrapper anyway to make the models consistently named

MartinHjelmare:
I'll make some refactors of the firmware models as a prerequisite to this PR to make it easier to use this in Home Assistant. There are many parts of the node and driver firmware models that are common and we should consolidate those first.

MartinHjelmare:
I've rebased this PR now. The scope is a lot smaller now after the firmware model refactor that is now merged.

zwave_js_server/model/driver/__init__.py
```diff
@@ -242,6 +249,38 @@ async def async_install_config_update(self) -> bool:
...
+            )
+        if update_data is not None and update_info is not None:
+            raise ValueError(
+                "Only one of update_data or update_info can be provided for firmware update."
+            )
+        params: FirmwareUpdateDataDataType | FirmwareUpdateInfoDataType
+        if update_data is not None:
+            params = update_data.to_dict()
+        elif update_info is not None:
+            params = update_info.to_dict()
```
AlCalzone, line 271: I think this is not correct. Z-Wave JS Server expects a property `updateInfo`:
https://github.com/zwave-js/zwave-js-server/pull/1432/files#diff-03515d137479c0dec74f922b42e9fb01e6828ea72077e82b07e0f919efae015aR103

but if I understand it correctly, this call spreads the child properties of that dict instead:
https://github.com/home-assistant-libs/zwave-js-server-python/pull/1262/files#diff-fffbb910cbecd6ea35b7baae1173e477ba1218a3def64d1aa3f045762d9ebbe1R65-R78

AlCalzone (changes requested):

AlCalzone (approved):

PR #1262, opened 2025-07-26
Refactor firmware models

MartinHjelmare:
- Consolidate common firmware model parts into base models.
- No functional changes.

### DIFF
--- a/zwave_js_server/model/node/firmware.py
+++ b/zwave_js_server/model/node/firmware.py
@@ -2,52 +2,56 @@
 
 from __future__ import annotations
 
-from dataclasses import asdict, dataclass, field
+from dataclasses import dataclass, field
 from enum import IntEnum
-from typing import TYPE_CHECKING, Literal, TypedDict, cast
+from typing import TYPE_CHECKING, Required, TypedDict, cast
 
-from ...const import VALUE_UNKNOWN, RFRegion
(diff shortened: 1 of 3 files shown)


AlCalzone (commented):
If I understand correctly, the number of files is not exposed for the driver firmware upgrade progress. Even though this is correct now, I wonder if we should expose that information too to stay future-proof without having to touch this again. Just in case zwave-js gains support for multi-step firmware updates for adapters.

Other than that LGTM.

zwave_js_server/model/node/firmware.py
```diff
@@ -133,193 +137,125 @@ class NodeFirmwareUpdateStatus(IntEnum):
...
 
 @dataclass
-class NodeFirmwareUpdateProgress:
+class NodeFirmwareUpdateProgress(FirmwareUpdateProgress):
     """Model for a node firmware update progress data."""
 
     node: Node
     data: NodeFirmwareUpdateProgressDataType = field(repr=False)
     current_file: int = field(init=False)
     total_files: int = field(init=False)
```
MartinHjelmare, line 154: Do you mean the attributes `current_file` and `total_files`?

If it's not supported now, I don't think we should add it. There's no problem with adding them later when they are supported by driver updates.

AlCalzone (approved):

PR #1417, opened 2026-04-14
Add support for schema 47 controller/node commands/events (3/4)

raman325:
## Summary

Third of four PRs adding schema 47 support, mirroring upstream zwave-js-server PR [#1513](https://github.com/zwave-js/zwave-js-server/pull/1513).

### Controller commands (5 new)

| Category | Commands |
|---|---|
| Diagnostics | `get_background_rssi`, `get_long_range_nodes` |
| Associations | `get_all_association_groups` , `get_all_associations` |
| Firmware | `get_all_available_firmware_updates` |

### Events (3 new)

- Controller: `joining network show dsk` (with `dsk` payload), `joining network done`
- Node: `check link reliability progress` (with `progress` dict)

### Strong typing

Command return types use typed models rather than raw dicts:

- **`BackgroundRSSI`** frozen dataclass for structured RSSI returns
- **`ZWaveChipType`** frozen dataclass (from PR 1/4) wrapping the `str | {type, version}` wire union

### Returns using `Node` objects

Methods that return node references resolve IDs to `Node` objects via `self.nodes`:
- `async_get_long_range_nodes` → `list[Node]`

## Test plan

- [x] `pytest -q` — all tests passing
- [x] black + ruff + pylint + mypy all pass
- [ ] CI green

🤖 Generated with [Claude Code](https://claude.com/claude-code)


zwave_js_server/model/controller/__init__.py
```diff
@@ -914,6 +914,582 @@ async def async_is_firmware_update_in_progress(self) -> bool:
...
+    async def async_get_dsk(self) -> bytes:
+        """Send getDSK command to Controller."""
+        data = await self.client.async_send_command(
+            {"command": "controller.get_dsk"},
+            require_schema=47,
+        )
+        return convert_base64_to_bytes(data["dsk"])
+
+    async def async_get_nvm_id(self) -> dict:
+        """Send getNVMId command to Controller."""
```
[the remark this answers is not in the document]
raman325: Fixed in #1422 — return type changed from `dict` to `int` to match the actual `data["nvmId"]` return value.

zwave_js_server/model/controller/__init__.py
```diff
@@ -710,15 +714,15 @@ async def async_remove_node_from_all_associations(
...
+    async def async_get_node_neighbors(self, node: Node) -> list[Node]:
         """Send getNodeNeighbors command to Controller to get node's neighbors."""
         data = await self.client.async_send_command(
             {
                 "command": "controller.get_node_neighbors",
                 "nodeId": node.node_id,
             }
         )
-        return cast(list[int], data["neighbors"])
+        return [self.nodes[nid] for nid in data["neighbors"]]
```
MartinHjelmare: Why do we change the return value of the existing methods? What does this has to do with schema 47?

test/model/test_controller.py
```diff
@@ -2441,3 +2456,720 @@ async def test_schema_47_network_lifecycle_events(controller: Controller) -> Non
...
+) -> None:
+    """Test assign custom return routes with and without priority route."""
+    ack_commands = mock_command(
+        {"command": "controller.assign_custom_return_routes"},
+        {"success": True},
+    )
+    routes = [Route(repeaters=[multisensor_6, multisensor_6], route_speed=100)]
+
+    result = await controller.async_assign_custom_return_routes(
+        multisensor_6, multisensor_6, routes
```
MartinHjelmare: It would be easier to understand the test and calls if we use keywords. Let's do that consistently if any parameter is passed without using an argument stored in a variable with the same name as the parameter, unless it's otherwise clearly explained.

Eg here it's easy to understand the `routes` parameter since we use a variable with that name. But it's harder to understand what the first and second parameters do since we pass the same node instance for both parameters.

test/model/test_controller.py
```diff
@@ -2441,3 +2456,720 @@ async def test_schema_47_network_lifecycle_events(controller: Controller) -> Non
...
+    method: str,
+    args: tuple[Any, ...],
+    response: dict[str, Any],
+    expected_none: bool,
+) -> None:
+    """Test cached route queries that may return None."""
+    mock_command({"command": f"controller.{wire_cmd}"}, response)
+    actual_args = tuple(multisensor_6 if a == "node" else a for a in args)
+    result = await getattr(controller, method)(multisensor_6, *actual_args)
+    if expected_none:
```
MartinHjelmare: Please avoid conditions in tests. Add/change parameters to avoid it or split the test.

zwave_js_server/model/controller/__init__.py
```diff
@@ -914,6 +918,596 @@ async def async_is_firmware_update_in_progress(self) -> bool:
...
+                "filterSubsets": filter_subsets,
+            },
+            require_schema=47,
+        )
+        return data.get("regions")
+
+    async def async_query_supported_rf_regions(self) -> list[int]:
+        """Send querySupportedRFRegions command to Controller."""
+        data = await self.client.async_send_command(
+            {"command": "controller.query_supported_rf_regions"},
```
MartinHjelmare: What's the difference between getting and querying the rf regions?

MartinHjelmare (commented):
I'm surprised we added all these things to schema 47. Most of them seem low level and not interesting for normal client operation and usage. I don't think we should add things just because we can. Did someone ask for these things?

zwave_js_server/model/controller/__init__.py
raman325: I was looking for consistency across schema 47. Where possible we should use the internal model, and when I asked Claude to review 1/4 and 2/4 PRs it came up with this. We don't have to make this change

raman325:
> I'm surprised we added all these things to schema 47. Most of them seem low level and not interesting for normal client operation and usage. I don't think we should add things just because we can. Did someone ask for these things?

for context, schema 47 was a realignment of the API between zwave-js and zwave-js-server. I don't expect us to use all of these methods, so the only risk here is changes to the API from pre schema 47. I think schema 48 is the big one because it changes how we access lock data and perform lock actions. That will require changes in HA but schema 46/47 probably won't unless we find a better method introduced for something we were already trying to get earlier - I don't think any of those exist

zwave_js_server/model/controller/__init__.py
MartinHjelmare: Please break it out to a separate PR as it's a breaking change.

MartinHjelmare:
But why did we add all the things to the server in schema 47? Did someone ask for them?

raman325:
No, schema 47 was meant to be a realignment - I found methods that were undocumented and were therefore missing from the server. We have modeled every command we've built into the server in this lib hence the PRs. I think this came from missing an event handler, and when I did a full sweep I found other things we were missing.

MartinHjelmare:
I don't think we should add features to the server just because they exist in the driver. There should be a need stated first.

Now we have added a score of features that may not be used and we have to support them forever both in the server and in the client.

raman325:
the server isn't just an interface for the lib, it's an interface for e.g. the .net version of zwave-js. We can opt not to support everything in the lib, but we've mirrored the zwave-js API as much as we can in the server. There are many commands the server supports as does the lib but aren't actively being used by the HA integration, so I am not sure what makes any of these different.

MartinHjelmare:
I know there are other clients. Again, my point is that we shouldn't add things just because it's in the driver. Low level things like these route setting commands we should not add unless someone explicitly asks for them.

zwave_js_server/model/controller/__init__.py
raman325: yes this is much older than any of those PRs good catch

zwave_js_server/model/controller/__init__.py
```diff
@@ -914,6 +919,107 @@ async def async_is_firmware_update_in_progress(self) -> bool:
...
+    ) -> dict[Node, dict[int, dict[int, list[AssociationAddress]]]]:
+        """Send getAllAssociations command to Controller."""
+        data = await self.client.async_send_command(
+            {
+                "command": "controller.get_all_associations",
+                "nodeId": node.node_id,
+            },
+            require_schema=47,
+        )
+        result: dict[Node, dict[int, dict[int, list[AssociationAddress]]]] = {}
```
MartinHjelmare: If we want to look up a value in the dict it may be inconvenient to need the node instance to do that.

I think we should consider using the node id as key instead.

raman325: Good point — reverted in 0ee5e44. The outer dict is now keyed by `int` (raw node IDs) so consumers can look up by ID directly.

MartinHjelmare (approved):

PR #1430, opened 2026-05-06
Add AI instructions

AlCalzone:
These have been compiled from recent review comments.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,45 @@
+# GitHub Copilot & Claude Code Instructions
+
+This repository is `zwave-js-server-python`, a Python client library for [zwave-js-server](https://github.com/zwave-js/zwave-js-server/). It mirrors the structure and events of Z-Wave JS 1:1 (`Driver`, `Controller`, `Node`, etc.).
+
+## Scope
+
+- One concern per PR. No bundling refactors, return-type changes, or renames into a feature/schema PR. Breaking changes go in their own PR.
+
+## Schema version
+
+- Tests and fixtures reference `MAX_SERVER_SCHEMA_VERSION` / `MIN_SERVER_SCHEMA_VERSION`. Never hardcode the number.
+- Helpers that send `schemaVersion` (e.g. `dump-state`) negotiate `min(MAX_SERVER_SCHEMA_VERSION, server_max_schema_version)`. No raw `MAX_SERVER_SCHEMA_VERSION`.
+
+## Typing
+
+- Return annotation matches actual return value. If body returns `data["nvmId"]` (`int`), annotate `int`, not `dict`.
+- No `Union` return when one type covers both known/unknown cases. Pick a type callers don't need to `isinstance`-discriminate.
+- JSON keys are `str`. If model annotates `dict[int, ...]`, convert keys at construction (dict comprehension). Don't lie in the annotation.
+- Public dict keyed by raw id (`int`), not model instance. Caller may not have the instance.
+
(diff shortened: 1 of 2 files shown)


AGENTS.md
```diff
@@ -0,0 +1,45 @@
...
+- `@cached_property` returns immutable container (`tuple`, not `list`).
+- New `@cached_property` requires `update()` to invalidate it, plus regression test: populate cache → `update()` → assert new value.
+- No per-call `dict(entry)` in cached/parsing methods. Use `typing.cast` or narrow with `in` check.
+
+## Match existing patterns
+
+Look for prior art in `zwave_js_server/model/` before inventing. Examples reviewers have flagged:
+
+- `from_dict` classmethod parses server payload: `def from_dict(cls, data: <Type>DataType) -> Self`. Returns `Self`, not the explicit class.
+- Rename unused parameters to `_param` (e.g. `event` in event handlers).
```
MartinHjelmare, line 33: I don't understand this. We don't have any parameters called `_param`.

AGENTS.md
```diff
@@ -0,0 +1,45 @@
...
+- No mutable module-level default on dataclass field. Copy in `__post_init__` (`list(DEFAULT_X)`).
+- `@cached_property` returns immutable container (`tuple`, not `list`).
+- New `@cached_property` requires `update()` to invalidate it, plus regression test: populate cache → `update()` → assert new value.
+- No per-call `dict(entry)` in cached/parsing methods. Use `typing.cast` or narrow with `in` check.
+
+## Match existing patterns
+
+Look for prior art in `zwave_js_server/model/` before inventing. Examples reviewers have flagged:
+
+- `from_dict` classmethod parses server payload: `def from_dict(cls, data: <Type>DataType) -> Self`. Returns `Self`, not the explicit class.
```
MartinHjelmare, line 32: This is just best practice Python type annotation using modern Python.

AGENTS.md
AlCalzone: The auto-generated instructions specifially mentioned `_event` but I wanted to keep it more generic. If it only ever comes up in event handlers, I can go back to being more specific.

AGENTS.md
AlCalzone: Right, but you flagged it during your most recent review. Opus 4.7 didn't follow best practices, so I think it needs a nudge in the right direction.

AGENTS.md
MartinHjelmare: I don't like this. We should keep the interface parameters always named the same. I suggest we remove this instruction and adjust the parameter names instead.

AGENTS.md
MartinHjelmare: Yes. Would it work to tell it to follow the latest type annotation best practices (proposed in PEPs) of the minimum supported version of the library?

`typing.Self` was added in Python 3.11. We support Python 3.12+.

https://docs.python.org/3/library/typing.html#typing.Self

https://peps.python.org/pep-0673/

AlCalzone: @abmantis do you have any experience with this approach?

abmantis: Its hard to say. Since its already "old" (3.11) it may work, but it is probably going to be flaky/inconsistent since its very broad.

As a side note: don't we have mypy/pyright/ruff/... rules we can use for these?

AlCalzone: > don't we have mypy/pyright/ruff/... rules we can use for these?

That would probably be the more reliable option.

PR #1431, opened 2026-05-06
Bump version to 0.71.0

MartinHjelmare:


raman325:
is the intention to keep the old utils for locks for at least one release? If not then that is still missingi from this bump

MartinHjelmare:
I think we can do that change separately. It's not dependant on the addition we've made already.

erwindouna (approved):

abmantis (approved):

PR #1442, opened 2026-06-01
Add support for schema 49, add_user command

AlCalzone:
Implements the new command exposed in https://github.com/zwave-js/zwave-js-server/releases/tag/3.9.0, required to fix https://github.com/home-assistant/core/issues/172466

### DIFF
--- a/zwave_js_server/const/__init__.py
+++ b/zwave_js_server/const/__init__.py
@@ -11,7 +11,7 @@ PACKAGE_NAME = "zwave-js-server-python"
 __version__ = "0.71.0"
 
 # minimal server schema version we can handle
-MIN_SERVER_SCHEMA_VERSION = 47
+MIN_SERVER_SCHEMA_VERSION = 49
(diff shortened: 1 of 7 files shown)


zwave_js_server/const/__init__.py
```diff
@@ -13,7 +13,7 @@
 # minimal server schema version we can handle
 MIN_SERVER_SCHEMA_VERSION = 47
```
MartinHjelmare: Should we bump the minimum schema version to 49? Otherwise some users with affected locks won't have working credential management, right?

AlCalzone: Sounds reasonable. I do need to release a version of the app with that schema first though.

AlCalzone: --> https://github.com/home-assistant/addons/pull/4641

MartinHjelmare:
Some tests need to be updated it seems.

MartinHjelmare (approved):
Thanks!

PR #1458, opened 2026-06-22
Forward the `interview progress` event

AlCalzone:
Adds support for the new `interview progress` event included in schema 50 (https://github.com/zwave-js/zwave-js-server/pull/1562).

I opted not to bump the minimum version, so we don't force users to upgrade for a nice-to-have UI improvement.

### DIFF
--- a/test/model/test_node.py
+++ b/test/model/test_node.py
@@ -1227,6 +1227,23 @@ async def test_interview_events(multisensor_6):
     assert node.awaiting_manual_interview
 
     event = Event(
+        type="interview progress",
+        data={
+            "source": "node",
+            "event": "interview progress",
+            "nodeId": 52,
+            "stage": "CommandClasses",
+            "progress": 42.5,
+            "endpoint": 0,
+            "commandClass": 112,
+        },
+    )
+    node.receive_event(event)
+    assert node.interview_stage == "CommandClasses"
+    assert not node.ready
+    assert node.in_interview
+
+    event = Event(
         type="interview stage completed",
         data={
             "source": "node",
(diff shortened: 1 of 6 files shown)


MindFreeze (approved):

test/model/test_node.py
```diff
@@ -1226,6 +1226,24 @@ async def test_interview_events(multisensor_6):
...
+            "source": "node",
+            "event": "interview progress",
+            "nodeId": 52,
+            "stage": "CommandClasses",
+            "progress": 42.5,
+            "endpoint": 0,
+            "commandClass": 112,
+        },
+    )
+    node.receive_event(event)
```
MartinHjelmare: We should assert something after this call so that the test would fail if we would remove the event handler.

MartinHjelmare (changes requested):

test/model/test_node.py
AlCalzone: The test now calls `handle_interview_progress` directly, just like we already do for `handle_interview_stage_completed` below.

test/model/test_node.py
```diff
@@ -1226,6 +1226,24 @@ async def test_interview_events(multisensor_6):
...
+            "source": "node",
+            "event": "interview progress",
+            "nodeId": 52,
+            "stage": "CommandClasses",
+            "progress": 42.5,
+            "endpoint": 0,
+            "commandClass": 112,
+        },
+    )
+    node.handle_interview_progress(event)
```
MartinHjelmare: Revert to use `receive_event` and assert that this debug log isn't logged.

https://github.com/home-assistant-libs/zwave-js-server-python/blob/94b43672215675609e693fd147c8efa997bbce00/zwave_js_server/event.py#L80

Also assert that the node instance is set in the event data.

MartinHjelmare (approved):
Thanks!

PR #1488, opened 2026-08-17
Fix wrong field name in toggle RF command

MindFreeze:
`async_toggle_rf` sends the boolean as `enable`, but zwave-js-server's handler reads `message.enabled` ([message_handler.ts](https://github.com/zwave-js/zwave-js-server/blob/master/src/lib/controller/message_handler.ts)), so `toggleRF` has always received `undefined`. Send `enabled` and rename the parameter to match.

Needed for home-assistant/core#179363.


MartinHjelmare (approved):
Thanks!

PR #1499, opened 2026-09-01
Exempt Basic CC from local value cache check in async_set_value

rahatarmanahmed:
In Z-Wave, all actuator devices support Basic CC (0x20), but node-zwave-js intentionally hides Basic CC targetValue from node.values when higher-level command classes (such as Binary Switch or Multilevel Switch) exist to avoid redundant entities.

Previously, Node.async_set_value enforced a local cache check that rejected any value ID not present in self.values with NotFoundError before dispatching the command to the driver, preventing setting Basic CC targetValue on devices exposing higher-level CCs.

Exempt CommandClass.BASIC from the self.values presence check and allow setting Basic CC targetValue on any node.

Fixes https://github.com/home-assistant/core/issues/128816

### DIFF
--- a/test/model/test_value.py
+++ b/test/model/test_value.py
@@ -14,6 +14,7 @@ from zwave_js_server.model.value import (
     ValueDataType,
     ValueMetadata,
     get_value_id_str,
+    parse_value_id_str,
 )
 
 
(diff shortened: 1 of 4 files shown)


test/model/test_value.py
```diff
@@ -336,3 +337,52 @@ def test_configuration_value_metadata(inovelli_switch_state):
     assert metadata.no_bulk_support
     assert metadata.value_size == 1
     assert metadata.format == ConfigurationValueFormat.SIGNED_INTEGER
+
+
+def test_parse_value_id_str(inovelli_switch):
```
MartinHjelmare: Please add complete type annotations to all new tests.

zwave_js_server/model/value.py
```diff
@@ -153,6 +153,38 @@ def get_value_id_str(
     return value_id
 
 
+def parse_value_id_str(node: Node, value_id: str) -> dict[str, Any] | None:
```
MartinHjelmare: This could be a protected helper method on the node instead.

MartinHjelmare (changes requested):

zwave_js_server/model/value.py
rahatarmanahmed: Moved over.

test/model/test_value.py
rahatarmanahmed: Done, I think. I'm new to types in python; I assume it's just the input arguments?

MartinHjelmare (approved):
Thanks!

PR #1507, opened 2026-09-19
Allow a null maxDataRate in the node ready event

Dextheking1:
## What does this change?

`NodeDataType.maxDataRate` was declared as `int`, so a node that sends an explicit `null` for it in the `ready` event made pydantic raise a `ValidationError` while validating the event in `ReadyEventModel.from_dict`. That exception propagates out of `Client.receive_until_closed()` and kills the client listen loop, which takes down the whole Z-Wave JS integration in Home Assistant: every Z-Wave entity becomes unavailable and the config entry needs a full restart to recover.

The field is now `int | None`, which also matches the existing `NodeModel.max_data_rate` accessor that already returns `int | None`.

Fixes home-assistant/core#182649

## Why?

Some nodes report `maxDataRate: null` (dead nodes, unknown status). A single node reporting null should not be able to take down the integration.

## Testing

Added `test_node_ready_event_null_max_data_rate`, which feeds a `ready` event whose `nodeState` carries an explicit `maxDataRate: null`. It fails on the previous code with the `ValidationError` from the report and passes with this change. The full suite (369 tests) passes, and ruff is clean.


brianfeucht (approved):
👍🏽

